PR #1, opened 2015-08-23
Install java and leiningen.

wkf:
Compiling sites built with [Aviary](https://github.com/wkf/aviary) is my primary use case, but this should enabling building most of the Clojure static site generators I've seen (since they use `lein`).


wkf:
The `apt-get` package in ubuntu is for leiningen 1, which is why this PR doesn't use it. 

Also, the `default-jdk` will install OpenJDK 7, which should be just fine.

wkf:
Oh, and dependencies end up in ~/.m2

PR #3, opened 2015-12-12
Install pandoc

benjaminnoury:
Install Pandoc universal document converter


benjaminnoury:
(pandoc-citeproc citation system is not included)

PR #7, opened 2016-07-11
Add git-lfs support

hhsnopek:


calavera:
Hi @hhsnopek. Thanks for helping out. I'm going to merge this and will deploy the image on our servers soon.

I just want to make sure you understand a couple of things:

1. running `git lfs install` at the image level won't configure your repository's filters. When you deploy your site on Netlify, you'll still need to run it inside the build process.

2. we don't have access to private repository information inside our containers (yet). You can work around that by setting an access token as environment variable in your settings.

PR #10, opened 2016-08-02
Install git-lfs from the official PPA.

calavera:


calavera:
This will unfortunately invalidate docker's cache and take forever to build. But it's the right way to add git-lfs.

PR #13, opened 2016-08-03
Add Git LFS filters and fetching rules.

calavera:
Allow to only download things that web sites use the most.
That way the cache that we generate is not full of assets that we're not going to serve.

See git-lfs-fetch for more information about include rules:

https://github.com/github/git-lfs/blob/7472995bcf0900c79ec6ae2d1fe86610b599744b/docs/man/git-lfs-fetch.1.ronn#L37


rybit:
lgtm

calavera:
anything else we should add to the filter?

calavera:
I guess we can add more later if people need them.

PR #16, opened 2016-08-09
add additional clean up for apt-get

hhsnopek:


calavera:
Can you add that command to the previous `RUN`, so we don't create an extra layer?

```
 RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && apt-get autoremove -y
```

hhsnopek:
def - fixed 👍 🎉

calavera:
Thanks! LGTM

PR #18, opened 2016-08-21
upgrade node to 6

kentcdodds:
I just did this in the github editor to get the conversation started (if it's not happening already).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -108,7 +108,7 @@ USER root
 USER buildbot
 RUN git clone https://github.com/creationix/nvm.git ~/.nvm
 
-RUN /bin/bash -c '. ~/.nvm/nvm.sh && nvm install 4 && nvm use 4 && \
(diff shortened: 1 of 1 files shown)


biilmann:
We are planning to make node 6 the default, but in any case we've adapted our build script so you can use any node version.

You can set the node version for the build, either by including an `.nvmrc` in your repository or by setting a `NODE_VERSION` environment variable (any value that nvm understands will work).

kentcdodds:
Great to know about the `NODE_VERSION` functionality 👍

biilmann:
Yes - fairly new addition (before you had to add a .nvmrc and depended on the versions available on the build image).

We still need to do a much better job of surfacing that option in the UI/Docs...

calavera:
We're going to make 6 the default, but we need to change a couple of things. I've opened #26 that also installs node 4 but makes node 6 the default and install some additional packages, like Yarn. I'm closing this PR since the problem will be solved in #26.

Thanks! :raised_hands:

PR #21, opened 2016-09-27
fix local build script

rybit:


run-build.sh
```diff
@@ -30,18 +30,17 @@ else
...
-		echo "Set node from .nvmrc"
-		export NODE_VERSION=$(cat .nvmrc)
-	else
-		echo "Error setting node version from .nvmrc file. Unsupported version?"
-		echo "Will use default version ($NODE_VERSION)"
-	fi
+source $HOME/.nvm/nvm.sh
+export NODE_VERSION=4
+if [ -f .nvmrc ]; then
+  nvm install $(cat .nvmrc)
```
calavera: do we want to run `nvm install` if the versions are the same?

calavera: I'm not sure if that's very efficient. Does `nvm install` do nothing if the versions are the same?

rybit: yeah it is a noop if it is already installed

from the image
```
buildbot@d52837537afd:/$ nvm install 4
v4.5.0 is already installed.
Now using node v4.5.0 (npm v2.15.9)
```

nicklasmoeller: When changing nvm version, your global node_modules aren't accessible from the previous version. So grunt, bower, etc. that you install [in the Dockerfile](https://github.com/netlify/build-image/blob/ccf8dbb8a5c25e58154cd61f0b56534670751242/Dockerfile#L111-L112) is not accessible, since the link to your node/bin dir is changed as well.

rybit: Yeah those should be in the package.json (better to pin your version). The those get added to your path [here](https://github.com/netlify/build-image/blob/master/run-build.sh#L86).

PR #23, opened 2016-10-12
Install Yarn and Elm in the base build image.

calavera:
So people can use Yarn to install packages and we can also build Elm projects.

We still need to check if our cache will work as expected with Yarn, but we'll need to do this anyways.


rybit:
lgtm

PR #30, opened 2016-11-01
Safety First - wrap all conditionals in double brackets

hhsnopek:
I also noticed that there's no style precedence set in this repo.


calavera:
LGTM, thanks!

PR #32, opened 2016-11-14
Add strace to the base image.

calavera:
So we can debug weird network issues with yarn.


rybit:
:shipit: (late but worth it!)

PR #34, opened 2016-11-18
Update pre-installed Rubies

t-richards:
The [Ruby downloads page](https://www.ruby-lang.org/en/downloads/) lists the latest supported versions of Ruby:

>#### Stable releases:
> - Ruby 2.4.1
> - Ruby 2.3.4
>
>#### In security maintenance phase (will EOL soon!):
> - Ruby 2.2.7
>
>#### Not maintained anymore (EOL):
> - Ruby 2.1.10

This PR:
 - Drops `2.0.x` and `2.1.x`
 - For each supported branch, adds the latest stable Ruby, and the `current_version - 1`.
 - Sets the RVM "default" Ruby to `2.2.7` (the oldest version which is still officially supported)
 - Allows for other non-preinstalled Rubies to be installed on the fly at site build time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -98,14 +98,13 @@ RUN gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys D39DC0E3 && \
 ENV PATH /usr/local/rvm/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
 RUN /bin/bash -c "source ~/.rvm/scripts/rvm && \
-                  rvm install 2.1.9 && rvm use 2.1.9 && gem install bundler && \
-                  rvm install 2.1.10 && rvm use 2.1.10 && gem install bundler && \
-                  rvm install 2.2.5 && rvm use 2.2.5 && gem install bundler && \
                   rvm install 2.2.6 && rvm use 2.2.6 && gem install bundler && \
-                  rvm install 2.3.2 && rvm use 2.3.2 && gem install bundler && \
+                  rvm install 2.2.7 && rvm use 2.2.7 && gem install bundler && \
                   rvm install 2.3.3 && rvm use 2.3.3 && gem install bundler && \
+                  rvm install 2.3.4 && rvm use 2.3.4 && gem install bundler && \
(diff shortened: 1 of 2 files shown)


biilmann:
Would love to merge this, but with the current setup we have to preserve the old ruby versions or we risk breaking people's build if they have .ruby-version files specifying one of these.

t-richards:
I see in `run-build.sh` that there's an on-the-fly nvm install: https://github.com/netlify/build-image/blob/22134a47d84338451c6f96dd59a3288197ea0a82/run-build.sh#L36

Would it be acceptable to do something similar for rvm?

calavera:
that should work, but it looks like there are some conflicts in the run-build.sh script. @t-richards do you mind resolving and pushing again?

t-richards:
@calavera Rebased against the latest `master`.

mitchellcash:
@calavera would you consider bumping the default version to 2.2.2?

The main reason I ask this is due to the fact the Ruby Gem [activesupport](https://rubygems.org/gems/activesupport/) has required Ruby version `>= 2.2.2` since June 30, 2016 (not considering Beta and RC releases).

Due to this being a reasonably popular gem (just under 113 million total downloads) it would be of assistance, rather than having to define a Ruby version in `/.ruby-version` for every project.

This Gem always seems to be the main reason why I have to end up defining a different Ruby version.

calavera:
@mitchellcash yes, I think we should make the current stable ruby version the new default. I need to find some time to test this properly and putting it in production without much hassle. Maybe keeping the current stable version in the image, so we can deploy it in two steps:

1. Update the image with all the new ruby versions installed.
2. Update the default ruby version we use.

mitchellcash:
@calavera thanks!

@t-richards and thanks for the quick updates to the pull request.

tysongach:
What’s the status on getting this merged?

t-richards:
So, it's April 2017 and 2.1.10 is not officially supported anymore. https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/

I'll bump versions again shortly.

mitchellcash:
@calavera still no movement here?

t-richards:
Will revisit at a later date.

PR #35, opened 2016-12-01
Update Yarn version to 0.17.10.

calavera:
- Includes a verbose flag and several improvements.
- Pin Elm to the current version we install, regardless their new
  release version, for now.


rybit:
lgtm

PR #36, opened 2016-12-13
fix busted symlink and update version

fool:
boot couldn't work as it was setup.  This fixes the symlink that was blocking it and updates to a newer version.


fool:
should notify https://app.intercom.io/a/apps/q245f50x/inbox/chrism@netlify.com/conversations/7186443864 when we *deploy* this

calavera:
LGTM

PR #41, opened 2017-01-12
Electron support through xvfb

schickling:
We're making heavy use of Electron (through [Nightmare](https://github.com/segmentio/nightmare))  for [build-time pre-rendering](https://github.com/graphcool/prep) and UI testing. Currently we're running all of our build code on Circle but it would be great if we could migrate more and more directly to Netlify this way.

To make Electron work inside of docker an additional dependency (`xvfb`) is required. You can find more information about this here: https://github.com/segmentio/nightmare/issues/224


schickling:
Thanks for merging this @biilmann 🎉

PR #43, opened 2017-01-18
Update YARN_VERSION

Leeds-eBooks:
Don't know if this is the way to do it! If not, consider this a request 🙂


rybit:
I think that the best way to do this is to actually set the environment variable `YARN_VERSION`. Updating and pushing a new build image is cumbersome and we tend to reserve it for structural changes b/c it can impact a lot of people. 

If that doesn't work feel free to ping us in support. I am going to close this for the time being.

PR #44, opened 2017-01-23
improve boot build speed (and hopefully fix mine)

naartjie:
The first thing `boot-clj` does when you invoke it, is it tries to update itself. This takes some time, and to do it on every run is not too efficient. I thought I would suggest the `boot -u` change to have all that dependency jazz in the base image.

Also my boot build is not working so I want to propose this to see if if helps any.


calavera:
LGTM

PR #49, opened 2017-03-11
Fixed Python 3 virtual environments

LenzGr:
Setting the Python version to `3.4.0` or `3.5.2` via `runtime.txt` did not work as documented on
https://www.netlify.com/docs/continuous-deployment/#set-node-ruby-or-python-version

The switch aborted with an error "Too many levels of symbolic links". Fixed the symlink creation which pointed back to itself instead the actual directory.

Signed-off-by: Lenz Grimmer <lenz@grimmer.com>


rybit:
good catch!

PR #51, opened 2017-03-26
Use binrc to install new Hugo versions on demand.

calavera:
Rather than adding every single version of Hugo to the base image,
from now on, users can install new versions of Hugo by setting
the HUGO_VERSION variable in their environments.

Binrc is a small program that will take that version and download the binary on demand. See more details in its repository:

https://github.com/netlify/binrc

Example: 

```
HUGO_VERSION 0.20 binrc install hugo
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -167,6 +167,12 @@ USER root
 #
 ################################################################################
 
+ENV BINRC_VERSION 0.1.1
+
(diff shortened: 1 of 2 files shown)


biilmann:
Awesome!

rybit:
lgtm ~ merging and starting to let it build

PR #52, opened 2017-03-27
install a bunch of language packages

### LINKED ISSUE #50: Add other locales to build environment so that system messages + number/date formatting can reflect different settings
Currently we only have a few locales, none non-US-english, installed in the build environment:

```
$ locale -a
C
C.UTF-8
en_US.utf8
POSIX
```

We have a request for French, but should probably just include all of the ones we can think of, since it's useful to people who aren't US-English speakers to be able to use the build environment in their native language and format things (like dates on blog posts) using system utilities.

rybit:
fixes #50 
and is in response to a customer request in intercom (for french, but while we're here...)


calavera:
Can you run `apt-get clean` after the install? That will remove cached files: `&& \
    apt-get clean`

https://github.com/netlify/build-image/blob/bf01a0892f101128001df857666f580e2f379a9c/Dockerfile#L24

calavera:
LGTM

PR #55, opened 2017-04-10
expose more tools for running builds

rybit:
This _should_ expose some helpful debugging scripts. They're aimed at making customer's lifes less difficult.


rybit:
cc/ @fool Could you double check that I am not lying in the README?

Issue #57, opened 2017-04-26
currently cannot use build image


fool:
`build X` does not work in the build-image:

```
$ docker run -it -v ${PWD}:/opt/repo netlify/build:latest /bin/bash
buildbot@a5628d0af185:/$ build hugo
Cloning into '/opt/buildhome/repo'...
done.
/usr/local/bin/build: line 27: /usr/local/bin/run-build-functions.sh: No such file or directory
Installing dependencies
/usr/local/bin/build: line 30: install_dependencies: command not found
Installing missing commands
/usr/local/bin/build: line 33: install_missing_commands: command not found
Executing user command: hugo
Caching artifacts
/usr/local/bin/build: line 40: cache_artifacts: command not found
buildbot@a5628d0af185:/$ find . -name run-build-functions.sh
find: `./etc/ssl/private': Permission denied
find: `./etc/polkit-1/localauthority': Permission denied
find: `./var/lib/php5': Permission denied
find: `./var/lib/polkit-1': Permission denied
find: `./var/cache/ldconfig': Permission denied
find: `./var/spool/cron/crontabs': Permission denied
find: `./var/spool/rsyslog': Permission denied
find: `./root': Permission denied
find: `./proc/tty/driver': Permission denied
buildbot@a5628d0af185:/$ exit
```

And, one cannot fix it:

```
$ cp ~/run-build-functions.sh .
$ git add run-build-functions.sh
$ git commit -m "add script"
[master 6782400] add script
 1 file changed, 2349 insertions(+)
 create mode 100644 run-build-functions.sh
$ docker run -it -v ${PWD}:/opt/repo netlify/build:latest /bin/bash
buildbot@624fc098b6cf:/$  cp /opt/repo/run-build-functions.sh /usr/local/bin
cp: cannot create regular file ‘/usr/local/bin/run-build-functions.sh’: Permission denied
```

speaking of this image:

`netlify/build                   latest              525b8d552832        25 hours ago        3.27 GB`

fool:
need to inform https://app.intercom.io/a/apps/q245f50x/respond/inbox/unassigned/conversations/9462202232 of any fix

calavera:
fixed in #60

fool:
@calavera @rybit I can't get this to work.  Please advise ASAP as customers also cannot get it to work:

```
$ docker pull netlify/build:latest
latest: Pulling from netlify/build
[...]

PR #58, opened 2017-04-26
make sure run-build-functions.sh is in the image

fool:
To address https://github.com/netlify/build-image/issues/57

Untested as yet


calavera:
Use the new script added in #60

PR #60, opened 2017-04-27
Add script to initialize a test container.

calavera:
This way we can control what's mounted on it
and we can set the volumes properly.


rybit:
yes. All the yes.

PR #62, opened 2017-05-01
mount the run-build-functions.sh as a file

rybit:
we were overwriting the /usr/local/bin/ when we just wanted to just add that file


calavera:
LGTM

PR #63, opened 2017-05-01
fixes for build image

rybit:
tested with out netlify-react-ui repo and can even set the YARN_VERSION and it works


calavera:
LGTM

PR #65, opened 2017-05-04
Up Yarn version to latest stable release

KyleAMathews:


rybit:
So you should be able to set this via the YARN_VERSION environment variable. As yarn moves quickly, we won't be updating the image constantly. 

https://github.com/netlify/build-image/blob/master/run-build-functions.sh#L52-L57 for reference. Feel free to ping me if that isn't true.

evenchange4:
Thank you, @rybit . It works! Netlify should add this useful information to documents. ✅

PR #71, opened 2017-07-12
Add jq to dockerfile

calebeby:
jq lets you parse json.

Example

```bash
curl https://my-blog-api.com/data | jq -r ".info"
```


calavera:
Do you mind fixing the conflict?

PR #73, opened 2017-07-23
Add Emacs Cask support

jcouyang:
- [Emacs](https://www.gnu.org/software/emacs/)
- [Cask](https://github.com/cask/cask)

As a Emacs user, I'm using [orgmode](http://orgmode.org/) to generate
my blog. And, I really like switch to netlify from github page
and circleci(emacs will compile orgmode files to html on the CI).

I've really like to have Emacs installed as part of netlify's build
tool(hugo can support partially orgmode parsing, but not native completely
support like emacs), and Cask as package management tool for Emacs.

So I can compile
Emacs orgmode files with command such as `cask exec emacs --batch -l
my-emacs-config.el -f org-publish-all` on netlify build, just like any ruby command `bundle
exec blah`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -251,14 +251,14 @@ RUN /bin/bash -c 'phpbrew init && source ~/.phpbrew/bashrc && phpbrew install 5.
 #
 ################################################################################
 USER root
-
-RUN wget http://ftpmirror.gnu.org/emacs/emacs-25.2.tar.gz && \
-    tar -xf emacs-25.2.tar.gz && \
-    cd emacs-25.2 && \
+ENV EMACS_VERSION 25.3
+RUN wget http://ftpmirror.gnu.org/emacs/emacs-${EMACS_VERSION}.tar.gz && \
+    tar -xf emacs-${EMACS_VERSION}.tar.gz && \
+    cd emacs-${EMACS_VERSION} && \
     env CANNOT_DUMP=yes ./configure --without-x && \
     make install && \
     make clean && \
-    cd .. && rm -rf emacs-25.2 emacs-25.2.tar.gz
+    cd .. && rm -rf emacs-${EMACS_VERSION} emacs-${EMACS_VERSION}.tar.gz
 
 USER buildbot
 RUN rm -rf /opt/buildhome/.cask && git clone https://github.com/cask/cask.git /opt/buildhome/.cask
--- a/Dockerfile
+++ b/Dockerfile
@@ -247,13 +247,19 @@ RUN /bin/bash -c 'phpbrew init && source ~/.phpbrew/bashrc && phpbrew install 5.
 
 ################################################################################
 #
-# Emacs Cask
+# Emacs and Cask
 #
 ################################################################################
 USER root
-RUN add-apt-repository ppa:kelleyk/emacs
-RUN apt-get -y update
-RUN apt-get -y install emacs25-nox
+
+RUN wget http://ftpmirror.gnu.org/emacs/emacs-25.2.tar.gz && \
+    tar -xf emacs-25.2.tar.gz && \
+    cd emacs-25.2 && \
+    env CANNOT_DUMP=yes ./configure --without-x && \
+    make install && \
+    make clean && \
+    cd .. && rm -rf emacs-25.2 emacs-25.2.tar.gz
+
 USER buildbot
 RUN rm -rf /opt/buildhome/.cask && git clone https://github.com/cask/cask.git /opt/buildhome/.cask
 ENV PATH "$PATH:/opt/buildhome/.cask/bin"
--- a/Dockerfile
+++ b/Dockerfile
@@ -245,9 +245,21 @@ USER buildbot
 RUN /bin/bash -c 'phpbrew init && source ~/.phpbrew/bashrc && phpbrew install 5.6 +default && \
     phpbrew app get composer'
 
+################################################################################
+#
+# Emacs Cask
+#
+################################################################################
 USER root
+RUN add-apt-repository ppa:kelleyk/emacs
+RUN apt-get -y update
+RUN apt-get -y install emacs25-nox
+USER buildbot
+RUN rm -rf /opt/buildhome/.cask && git clone https://github.com/cask/cask.git /opt/buildhome/.cask
+ENV PATH "$PATH:/opt/buildhome/.cask/bin"
 
 # Cleanup
+USER root
 RUN rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && apt-get autoremove -y
 
 # Add buildscript for local testing
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -326,6 +326,16 @@ install_dependencies() {
       export PATH=$(dirname $hugoPath):$PATH
     fi
   fi
+
+  # Cask
+  if [ -n "Cask" ]
+  then
+    cask install
+    if [ $? -eq 0 ]
+    then
+      echo "Emacs packages installed"
+    fi
+  fi
 }
 
 #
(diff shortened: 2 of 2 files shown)


kaushalmodi:
Can you also update this to install the latest version of stable Org?

Dockerfile
```diff
@@ -245,9 +245,21 @@ USER buildbot
...
 
+################################################################################
+#
+# Emacs Cask
+#
+################################################################################
 USER root
+RUN add-apt-repository ppa:kelleyk/emacs
+RUN apt-get -y update
+RUN apt-get -y install emacs25-nox
```
Benaiah: This line will have to be updated when Emacs gets past 25 (though that's admittedly a ways out at the moment). IMO it'd be better not to use a package with the major version hardcoded. Also, does this install the binary as `emacs` or `emacs25`?

Also, this is the `nox` version, which drops X dependencies. I'm not aware of any implications this would have on somebody using Emacs as a build tool, but this is not the most common build configuration. I think that's fine, since I can't imagine anybody needing graphical frames as part of an automated build, but I wanted to note that.

Dockerfile
```diff
@@ -245,9 +245,21 @@ USER buildbot
 RUN /bin/bash -c 'phpbrew init && source ~/.phpbrew/bashrc && phpbrew install 5.6 +default && \
     phpbrew app get composer'
 
+################################################################################
+#
+# Emacs Cask
+#
+################################################################################
 USER root
+RUN add-apt-repository ppa:kelleyk/emacs
```
kaushalmodi: > v25.2 isn't all that different from v24.4

v25.2 has a **lot** of changes since 24.4. See [`NEWS.25`][1].

[1]: http://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS.25

jcouyang:
@kaushalmodi 
> Can you also update this to install the latest version of stable Org?
org is a package, should be installed as package by user in [Cask file](https://github.com/jcouyang/blog/blob/master/Cask#L4)

just like ruby or any other lang, you define packages in Gemfile, not when you install ruby

Dockerfile
jcouyang: I'm not a ubuntu user, I'm happy to hear any suggest on how to install emacs 25 on ubuntu officially.

AFAICT ubuntu 14 officially only provide emacs 24.3, but most moden emacs packages require at lease 24.4.

CMIIW i think kelleyk/emacs just provide deb pacakge, it's still build from official emacs source

Benaiah: True - I meant changes that would affect using Emacs as a build tool, though that's admittedly hard to predict. If the 24->25 upgrade is really affecting build, would it be possible to use the `emacs25` package used in newer Ubuntu versions (see https://packages.ubuntu.com/zesty/editors/emacs)?

jcouyang: >  If the 24->25 upgrade is really affecting build

sorry I'm not following, this is not an upgrade, the Ubuntu 14 docker don't have any emacs preinstalled, how it's gonna affect v24 build?

Benaiah:
@kaushalmodi while Cask files are not what I'd personally recommend (it's infrequently updated, written in Python instead of Elisp for no real reason, and I don't have a great experience working with it), Emacs also has built-in packaging utilities, which include the ability to upgrade to a newer version of org. I personally use [use-package](https://github.com/jwiegley/use-package) with [quelpa](https://github.com/quelpa/quelpa) and [quelpa-use-package](https://github.com/quelpa/quelpa-use-package), but the packaging facilities built into Emacs are decent as well. I don't see a need to upgrade Org in the Dockerfile, particularly since people using the default version may see their scripts break (Org is notorious for breaking its API when updating).

Dockerfile
calavera: It looks like orgmode should work just fine with v24.4:

http://orgmode.org/Changes.html

> From now on, Org expects at least Emacs 24.3, although Emacs 24.4 or above is suggested.

I'd rather use the distro version if the library works the same. We don't need all the emacs features in this case.

Dockerfile
calavera: We want the `nox` version here.

Dockerfile
jcouyang: yeah, v24.4 would work just fine, but distro version is 24.3, which is too old and useless, even some common package like `scala-mode` or `clojure-mode` (for code highlight in org export) will not work.

to get 24.4 or above IMO there're 3 options:
1. ppa
2. build from source code(there's already [build-essentials](https://github.com/netlify/build-image/blob/master/Dockerfile#L12) installed, so I guess it would just work)
3. change the base image from Ubuntu 14 to 16, which have emacs 24.5 included(but could be big impact to other packages in build-image)

DamienCassou:
Do you know about [EVM](https://github.com/rejeep/evm)? Here is how to use it in travis:

```yaml
before_install:
  - curl -fsSkL https://gist.github.com/rejeep/ebcd57c3af83b049833b/raw > x.sh && source ./x.sh
  - evm install $EVM_EMACS --use --skip
  - cask

env:
  - EVM_EMACS=emacs-25.1-travis
  - EVM_EMACS=emacs-25.2-travis
  - EVM_EMACS=emacs-git-snapshot-travis
```

jcouyang:
@DamienCassou 
EVM looks awesome, but from it's source code, it still build from source, so I think build one on docker is more reasonable to me if ppa is a concern.

@calavera 
> I'd rather use the distro version if the library works the same

distro version is 24.3, most modern emacs packages require at lease 24.4

@Benaiah configure it as `without-x` we don't need x window for emacs batch mode

the environment variable `CANNOT_DUMP=yes` is because the issue https://github.com/moby/moby/issues/22801 but it doesn't affect anything in batch mode so I think that's fine for orgmode parser purpose here.

kaushalmodi:
I'd love to see Emacs being available on Netlify even more now that I have a stable blogging flow of exporting Hugo posts directly from Org (https://ox-hugo.scripter.co). 

Any update on this please?

kaushalmodi:
@Benaiah It would be nice if a minimal emacs binary is available on Netlify as we won't need GUI frames, etc. .. just enough to run it in batch mode, and load/run other libraries like org-mode.

Emacs is more than a text editor.. it's a lisp environment.

Benaiah:
@kaushalmodi to be clear - I'm totally in favor of having Emacs support (I've used it as an SSG myself in the past). The only major issue I originally had with this PR is that 24.4 (which is different from 24.3 in very significant ways) isn't available from the default Ubuntu repos for our version of Ubuntu, but I'm not comfortable building Emacs out of a random PPA. I suggested above (in the discussion on the review comments) that we see if we can get Emacs 25 by pulling it from the source of a newer Ubuntu version - that seems like the simplest option to me. Building from source (as the PR is currently doing) is also acceptable, but @calavera would have to chime in on how feasible that is within our infrastructure, as I'm not an expert on it.

Dockerfile
```diff
@@ -247,13 +247,19 @@ RUN /bin/bash -c 'phpbrew init && source ~/.phpbrew/bashrc && phpbrew install 5.
...
-# Emacs Cask
+# Emacs and Cask
 #
 ################################################################################
 USER root
-RUN add-apt-repository ppa:kelleyk/emacs
-RUN apt-get -y update
-RUN apt-get -y install emacs25-nox
+
+RUN wget http://ftpmirror.gnu.org/emacs/emacs-25.2.tar.gz && \
```
kaushalmodi: @Benaiah Looks like PPA is no longer used in the updated PR.

kaushalmodi: Sorry, so understood.. you don't have any issue with the updated PR, and we are waiting on feedback from @calavera. 

@calavera Please look at this PR.

calavera:
I'll test it out this week with a bunch of other updates we have in the queue to production.

kaushalmodi:
Very cool! While you are testing it, can you change that URL to get Emacs 25.3 instead of 25.2? @jcouyang Can you update the PR?

jcouyang:
Thanks @calavera, looking forward to hearing good news from you soon.
great idea, @kaushalmodi I just update and parameterize the emacs version so future upgrade will be easier as well

calavera:
This seems to be working on my tests. I'm going to merge it, but we won't deploy it just yet. I'll leave another comment here when that happens.

kaushalmodi:
@calavera Looks like now `emacs` available for a while on Netlify! I was waiting for a followup comment to happen here, and I just noticed that emacs from Netlify was being used instead of my version. :)

PR #75, opened 2017-08-02
start-image.sh: PWD -> pwd

thewtex:
Addresses PWD command not found on Ubuntu 16.04.

Fixes #74


rybit:
lgtm ~ shouldn't break other distros.

PR #83, opened 2017-09-20
Add Hugo 0.27.1

matt40k:
Add Hugo 0.27.1


calavera:
You can set HUGO_VERSION in your build environment and we'll use that version, but we should really do this. Thanks 🙌

matt40k:
Ah sweet!! Tried hugo_0.27.1 but that didn't work - helps if I read the doc lol!

> For version hugo > 0.20 you have to specify version hugo for testing and production in netlify.toml file or set HUGO_VERSION as a build environment variable in the Netlify console.

Is it worth also update the url so it refers to https://github.com/gohugoio/hugo rather then https://github.com/spf13/hugo - (ok the redirects are working so its not broken, but...)

fool:
updated that link just now, @matt40k .  Will close your PR since we don't want to add any more hugo versions that way ;)

PR #105, opened 2017-11-21
Remove unused command

aureliojargas:
The variables on the next two lines use absolute paths, so the current working directory, at this point, is not relevant.

And following that there is another `cd` command, now entering the real relevant directory.


brycekahle:
Thanks!

PR #107, opened 2017-11-21
Simplify logic using || instead of negated if

aureliojargas:
Shorter and easier to read.


aureliojargas:
Using the `test` command explicitly it would be even more readable, since the square brackets are gone:

```bash
test -f $1 || return 0
test -f $3 || return 0
```

calavera:
lgtm

PR #109, opened 2017-12-03
Add latest patch release for lowest supported Ruby (2.2) and default

mitchellcash:
Ruby version 2.1.x reached end of life in March 2017.

As it no longer receives security fixes my thoughts were to default to the lower major Ruby version still supported and the latest available patch. As of today this is 2.2.8.

Question outside of the scope of this pull request, does Netlify have a deprecation plan for 2.1.x and lower?


brycekahle:
> Question outside of the scope of this pull request, does Netlify have a deprecation plan for 2.1.x and lower

We don't have a specific plan for deprecating Ruby versions yet. Removing versions from the build image that users depend on is particularly impactful, in a bad way. Every build has to install that version of Ruby, which takes a long time.

If you have ideas on how to handle deprecations without a lot of user impact, we would love to hear them!

brycekahle:
Fixed by https://github.com/netlify/build-image/pull/131
Thanks!

mitchellcash:
@brycekahle nice work, thanks!

PR #115, opened 2017-12-14
add .emacs.d to cache so each build don't need to reinstall all those packages

jcouyang:
while #73 is working, we
also need to cache the packages `.emacs.d` if user install packages
with `package.el`.


brycekahle:
@jcouyang would you mind rebasing your PR? Thanks!

PR #120, opened 2017-12-26
Update libvips to 8.6

### LINKED ISSUE #42: Update libvips?
Any reason you're using such an old (2 years) version?

patrick91:
Fixes #42 and this error I'm having using `sharp`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,12 +69,12 @@ RUN wget https://www.python.org/ftp/python/3.6.2/Python-3.6.2.tar.xz && \
 
 WORKDIR /tmp
 
-# this actually builds v7.42.4 off of the 7.42 branch
+# this actually builds v8.6.2
 RUN \
-  curl -sLo libvips-7.42.zip https://github.com/jcupitt/libvips/archive/7.42.zip && \
-  unzip libvips-7.42.zip && \
-  cd libvips-7.42 && \
-  ./bootstrap.sh && \
+  curl -sLo vips-8.6.2.tar.gz https://github.com/jcupitt/libvips/releases/download/v8.6.2/vips-8.6.2.tar.gz && \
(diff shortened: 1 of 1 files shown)


patrick91:
@biilmann @brycekahle is it possible to merge this?

brycekahle:
@patrick91 It looks like based on the [instructions](https://github.com/jcupitt/libvips#building-libvips-from-a-source-tarball) we should be using the `.tar.gz` source tarballs now. The commands to build also need some updating. We may also be missing some of the dependencies for this version. 

Have you attempted to build the Dockerfile as you have it?

patrick91:
I think I did, I'll try again tomorrow, and update the commands as well while I'm at it :)

patrick91:
@brycekahle update with the released tar file, the docker build wen through, it just took a looong time :)

Dockerfile
```diff
@@ -69,12 +69,12 @@ RUN wget https://www.python.org/ftp/python/3.6.2/Python-3.6.2.tar.xz && \
...
+# this actually builds v8.6.2
 RUN \
-  curl -sLo libvips-7.42.zip https://github.com/jcupitt/libvips/archive/7.42.zip && \
-  unzip libvips-7.42.zip && \
-  cd libvips-7.42 && \
-  ./bootstrap.sh && \
+  curl -sLo vips-8.6.2.tar.gz https://github.com/jcupitt/libvips/releases/download/v8.6.2/vips-8.6.2.tar.gz && \
+  tar xvf vips-8.6.2.tar.gz && \
+  cd vips-8.6.2 && \
+  sh autogen.sh && \
```
brycekahle: I don't think this is necessary anymore, according to their docs. It was, only if we were cloning from git (or downloading the source as a zip as we were before).

PR #125, opened 2018-01-13
Remove EOL and only support latest patched Rubies

DirtyF:
> Ruby 2.2 is now under the state of the security maintenance phase, until the end of the March of 2018. After the date, maintenance of Ruby 2.2 will be ended. We recommend you start planning migration to newer versions of Ruby, such as 2.4 or 2.3. — https://www.ruby-lang.org/en/news/2017/12/14/ruby-2-2-9-released/

FWIW Jekyll silently dropped support for Ruby 2.1 and will drop support for 2.2 in next major version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -128,17 +128,11 @@ RUN gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys D39DC0E3 && \
 ENV PATH /usr/local/rvm/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
 RUN /bin/bash -c "source ~/.rvm/scripts/rvm && \
-                  rvm install 2.0.0-p247 && rvm use 2.0.0-p247 && gem install bundler && \
-                  rvm install 2.1.2 && rvm use 2.1.2 && gem install bundler && \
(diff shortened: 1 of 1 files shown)


brycekahle:
I appreciate the sentiment. Unfortunately if a user build specifies on a older version of Ruby that this PR removes, the build will have to install it. That process takes a long time, and will happen on each build.

We need to have a conversation about how to handle deprecating versions of Ruby with minimal impact. If you have ideas, we would to love to hear them.

DirtyF:
I'm no sysadmin, but I guess you'll have either to warn user in the build that he should update or redirect to a supported Ruby version?

calavera:
I think it'd be ok if we set RVM to automatically download versions when they are not installed and then cache them like we do with nvm versions. See the `Install on Use` section:

https://rvm.io/rubies/installing

brycekahle:
@calavera The major downside to that is Ruby installs take significant longer than Node. The cache only works for specific branches, so anytime you create a new branch you pay the Ruby install time penalty.

brycekahle:
We could use that time penalty as a nudge towards the latest patched versions in 2.2, 2.3, 2.4, and 2.5 though I suppose.

PR #126, opened 2018-01-22
Fallback to .node-version if .nvmrc isn't present

### LINKED ISSUE #79: respect .node-version if .nvmrc isn't there.
@biilmann suggested this was reasonable

I suggest something like `if [ ! .nvmrc ] ; then if [ .node-version ] use .node-version` type processing, so we don't make a breaking change to anyone using nvmrc and happening to have a conflicting .node-version file

levibuzolic:
Quick attempt to resolve #79 - allowing the use of `.node-version` files when `.nvmrc` isn't present.

https://twitter.com/fool/status/955586493312937984

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -157,6 +157,12 @@ install_dependencies() {
   then
     NODE_VERSION=$(cat .nvmrc)
     echo "Using node version '$NODE_VERSION' from .nvmrc"
+  else
+    if [ -f .node-version ]
+    then
+      NODE_VERSION=$(cat .node-version)
+      echo "Using node version '$NODE_VERSION' from .node-version"
+    fi
   fi
 
   if nvm install $NODE_VERSION


levibuzolic:
@brycekahle any chance this could get a set of :eyes: ?

brycekahle:
@levibuzolic What tooling uses a `.node-version` file?

levibuzolic:
n, avn, direnv, nodenv

run-build-functions.sh
```diff
@@ -157,6 +157,12 @@ install_dependencies() {
   then
     NODE_VERSION=$(cat .nvmrc)
     echo "Using node version '$NODE_VERSION' from .nvmrc"
+  else
+    if [ -f .node-version ]
```
brycekahle: Would you mind collapsing this `else` and `if` into an `elif` statement?

brycekahle:
Fixed by https://github.com/netlify/build-image/commit/3b563b15747e6e9e7bae7780e190db91e787b205

Thanks!

PR #130, opened 2018-02-05
Update build-image to match current defaults for new sites

### LINKED ISSUE #128: Upgrade to recent yarn by default
right now you can set YARN_VERSION as desired but then you will almost certainly clash with the `--ignore-optional` we have as default to cope with problems in our current version.

When updating, make sure we do something conditional to preserve the behavior of older builds pinned at yarn 0.18.1

brycekahle:
Fixes #128 

I'm curious if we should update Ruby as well, 2.2 is planned to be EOL on 2018-03-31.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -19,7 +19,7 @@ mkdir -p $NETLIFY_CACHE_DIR/.cask
 mkdir -p $NETLIFY_CACHE_DIR/.m2
 mkdir -p $NETLIFY_CACHE_DIR/.boot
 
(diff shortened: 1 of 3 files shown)


brycekahle:
For existing sites using `0.18.1` that need the `--ignore-optional` flag, we could set that env var for them in the database if they do not have an existing `YARN_FLAGS`.

rybit:
lgtm 

we need to do the database thingy first

brycekahle:
We decide to use a code-based approach in the API. No database changes required.

PR #131, opened 2018-02-05
Reduce installed ruby to latest supported versions. 

brycekahle:
Cache custom versions.

Fixes #125 #109


calavera:
@brycekahle I think we should only install the stable versions listed here:
https://www.ruby-lang.org/en/downloads/

brycekahle:
`2.2.9` is technically not EOL yet. That doesn't happen until 2018-03-31. I cannot install `2.5.0` yet until RVM does another release.

calavera:
> 2.2.9 is technically not EOL yet. That doesn't happen until 2018-03-31. I cannot install 2.5.0 yet until RVM does another release.

ugh, ok, I thought they had a binary already.

calavera:
ship it

mitchellcash:
@brycekahle appreciative of these changes, much more flexible overall and easier to manage from your perspective. When will this make it to Netlify production?

brycekahle:
@mitchellcash I'd say in the next day or two. I'll ping this issue when it is released.

brycekahle:
@mitchellcash The changes are now be live.

mitchellcash:
@brycekahle thanks for the notice!

PR #134, opened 2018-02-06
Add php 7.0 and 7.1 as selectable versions

jenschude:
PHP 5.5 already reached end of life.
PHP 5.6 and PHP 7.0 only provide security fixes support.
PHP 7.1 is in active support life cycle

Also more and more packages are in the phase of dropping 5.x support.


brycekahle:
It looks like 7.2 and 5.6 are the current stable versions. Any reason to include 7.0 and 7.1?

jenschude:
I can add the 7.2 also.

And there are some breaking changes between 7.0, 7.1 and 7.2. Also the possibility that someone runs into issues is very low, they could be added for completeness

jenschude:
Oh just saw you are now using the ubuntu alternatives instead of phpbrew. So this PR is no more relevant :)

PR #102, opened 2017-11-07
Add PHP and Composer support

Narno:
- switch to PHP 5.6(.32)
- install Composer and run `composer install`
- cache ~~`vendor` and~~ `.composer`
- download [PHPoole](https://phpoole.org).phar if necessary

PR #135, opened 2018-02-06
Add better PHP support

### LINKED ISSUE #124: better php support in build image
Ported from the feature requests listing, seems to belong here instead:

## Summary
We should support more php version the way we support node versions (with an easy config flag) and probably also some PHP dependency manager.

## Details

We currently support an ancient (but stock for our ubuntu release) php version, 5.5. Via phpbrew we kiiind of support php 5.6. It's not something anyone would figure out how to use without some guidance, but it is at least possible

The current version of PHP is 7.2 and while this is a fairly recent release, it's also a major change (there was no production-quality php 6 release).

We should also support a dependency manager like Composer (think Ruby bundler for php)

**Note:  there are very few PHP users on our platform, so we are unlikely to pursue this - great opportunity for a contribution if it's something you need!**

brycekahle:
Remove php 5.5, add 7.2.2
Automatically install composer dependencies and cache them.

Fixes #124
Fixes #102

Co-Authored-By: Arnaud Ligny <github@narno.org>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -356,8 +356,8 @@ RUN update-alternatives --set php /usr/bin/php5.6 && \
 
 USER buildbot
 
-RUN mkdir -p ~/.php && ln -s /usr/bin/php5.6 ~/.php/php
-ENV PATH "$HOME/.php:$PATH"
+RUN mkdir -p /opt/buildhome/.php && ln -s /usr/bin/php5.6 /opt/buildhome/.php/php
+ENV PATH "/opt/buildhome/.php:$PATH"
 
 ################################################################################
 #
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,41 +8,151 @@ MAINTAINER Netlify
 #
 ################################################################################
 
-RUN apt-get -y update && \
-    apt-get install -y build-essential g++ libssl-dev curl wget zip \
-                      apache2-utils libxml2-dev libxslt-dev python-setuptools \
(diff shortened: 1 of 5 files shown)


rybit (approved):
we will need to start passing the php version through to these commands.

brycekahle:
It is just another env var that can be set for a site and passed. No special support needed.

Narno:
Thanks for you great job @brycekahle! :+1: 

There is a way to contribute to the official documentation? https://www.netlify.com/docs/build-settings/#build-environment-variables

brycekahle:
@Narno not at the moment. Once this is live, I'll make sure our docs team makes the necessary changes.

minthemiddle:
@brycekahle How can I find out when it is live? Would be willing to write a tutorial on how to use it.

brycekahle:
@minthemiddle This is live now.

minthemiddle:
I just published “Just commit to git: Build static pages continuously using PHP generators like Jigsaw on Netlify” https://medium.com/@martin_betz/just-commit-to-git-build-static-pages-continuously-using-php-generators-like-jigsaw-on-netlify-4a3a18327ffa where I use all the latest improvements on PHP in the build image.

brycekahle:
Awesome. Thanks for the shout out!

nelson6e65:
There is a way to set environment variables to customize `composer install` args? For example, I need to use `--ignore-platform-req=php` in a specific site

PR #137, opened 2018-02-07
Add webp

monty5811:
Would it be possible to add the `webp` package to the build please?

I am currently using `cwebp` in my build step, but I have to download the binary each time.

(Or is there an easier way to do this, I couldn't figure out how to apt-get install during a build)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -137,6 +137,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         strace \
         swig \
         unzip \
+        webp \
         wget \
         xvfb \
         zip \


brycekahle:
@monty5811 We are currently exploring the idea of allowing you specify apt packages to install via the `netlify.toml` file. This feels like a better long-term solution for packages which may not have widespread usage. Individually they don't add much to the build image size, but together it is significant.

Can you explain your use case a little more? We are definitely open to including packages that would benefit a lot of our users.

monty5811:
Thanks for the reply. 

I was "shelling out" from python to call the `cwebp` binary to convert some images in my build step. 

Doing it in the build step means I only have to store the original JPGs.

I have since come up with a workaround: use Pillow from inside python to do the conversion (since  libwebp is already in the build image this works great).

So, I suppose my root issue is fixed, you could close this pull request, but someone else that wants to use cwebp in a language that doesn't have bindings may still find this change useful?

PS the apt install from the netlify.toml would be very cool.

brycekahle:
I'm going to close for now. If we get more requests for it, we can reconsider.

edmorley:
> We are currently exploring the idea of allowing you specify apt packages to install via the netlify.toml file.

What do you think about a generic `pre-build` command hook, that would allow people to both achieve the above (presuming the builds are run as root in the container; if not I guess you were meaning a TravisCI APT-addon like solution?), plus allow working around #141 or any other edge cases?

brycekahle:
Unfortunately, the user commands are not run as root. We were thinking more along the lines of the TravisCI apt solution, but we haven't settled on anything yet.

A pre-build is something we could consider too.

PR #160, opened 2018-03-05
Add support with Go through Gimme.

calavera:
- Install the most recent version and allow to install new versions on demand.
- Setup the project in the GOPATH when GO_IMPORT_PATH is set.
- Cache user versions.
- Cache user gopath.

This will allow us to support building Go functions for AWS lambda.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -452,16 +452,24 @@ install_dependencies() {
     local goVersion=$(cat .go-version)
     if [ "$installGoVersion" != "$goVersion" ]
     then
-	installGoVersion="$goVersion"
+      installGoVersion="$goVersion"
     fi
   fi
 
   if [ "$GIMME_GO_VERSION" != "$installGoVersion" ]
   then
-    echo "Installing Go version $goVersion"
-    eval "$(GIMME_ENV_PREFIX=$HOME/.gimme_cache/env GIMME_VERSION_PREFIX=$HOME/.gimme_cache/versions gimme $goVersion)"
+    echo "Installing Go version $installGoVersion"
+    GIMME_ENV_PREFIX=$HOME/.gimme_cache/env GIMME_VERSION_PREFIX=$HOME/.gimme_cache/versions gimme $installGoVersion
+    if [ $? -eq 0 ]
+    then
+      source $HOME/.gimme_cache/env/go$installGoVersion.linux.amd64.env
+    else
+      echo "Failed to install Go version '$isntallGoVersion'"
+      exit 1
+    fi
   else
-    eval "$(gimme)"
+    gimme
+    source $HOME/.gimme/env/go$GIMME_GO_VERSION.linux.amd64.env
   fi
 
   # Setup project GOPATH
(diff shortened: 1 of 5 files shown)


run-build-functions.sh
```diff
@@ -437,6 +444,33 @@ install_dependencies() {
...
   fi
+
+  # Go version
+  restore_home_cache ".gimme_cache" "go cache"
+  if [ -f .go-version ]
+  then
+    local goVersion=$(cat .go-version)
+    if [ "$installGoVersion" != "$goVersion" ]
+    then
+	installGoVersion="$goVersion"
```
brycekahle: weird indentation here

run-build-functions.sh
```diff
@@ -437,6 +444,33 @@ install_dependencies() {
...
+    fi
+  fi
+
+  if [ "$GIMME_GO_VERSION" != "$installGoVersion" ]
+  then
+    echo "Installing Go version $goVersion"
```
brycekahle: This should be `$installgoVersion` correct?

run-build-functions.sh
```diff
@@ -437,6 +444,33 @@ install_dependencies() {
...
+    eval "$(GIMME_ENV_PREFIX=$HOME/.gimme_cache/env GIMME_VERSION_PREFIX=$HOME/.gimme_cache/versions gimme $goVersion)"
```
brycekahle: Anyway we can avoid `eval`? Also, we usually fail the build if we fail to install the language version requested.

run-build-functions.sh
```diff
@@ -455,6 +489,15 @@ cache_artifacts() {
...
   cache_home_directory ".composer" "composer dependencies"
 
+
+  # Don't follow the Go import path or we'll store
+  # the origin repo twice.
+  if [ -n "$GO_IMPORT_PATH" ]
+  then
+    unlink $GOPATH/src/$GO_IMPORT_PATH
+  fi
+  cache_home_directory ".gimme_cache" "composer dependencies"
```
brycekahle: `composer` -> `go`

calavera:
@brycekahle comments addressed

run-build-functions.sh
```diff
@@ -452,16 +452,24 @@ install_dependencies() {
...
   then
-    echo "Installing Go version $goVersion"
-    eval "$(GIMME_ENV_PREFIX=$HOME/.gimme_cache/env GIMME_VERSION_PREFIX=$HOME/.gimme_cache/versions gimme $goVersion)"
+    echo "Installing Go version $installGoVersion"
+    GIMME_ENV_PREFIX=$HOME/.gimme_cache/env GIMME_VERSION_PREFIX=$HOME/.gimme_cache/versions gimme $installGoVersion
+    if [ $? -eq 0 ]
+    then
+      source $HOME/.gimme_cache/env/go$installGoVersion.linux.amd64.env
+    else
+      echo "Failed to install Go version '$isntallGoVersion'"
```
brycekahle: typo, should be `$installGoVersion`.

run-build-functions.sh
```diff
@@ -452,16 +452,24 @@ install_dependencies() {
...
+      exit 1
+    fi
   else
-    eval "$(gimme)"
+    gimme
```
brycekahle: Can this fail? If so, we should add a `if ! gimme, then, fi` block.

calavera: it should be a noop in network, but I'm sure it could fail in some other way. I added a check.

run-build-functions.sh
calavera: I added a check :raised_hands:

calavera:
Updated @brycekahle

PR #163, opened 2018-03-15
recently changed so you need both or grunt-cli won't "run"

fool:
we aren't sure when it changed, but the effect of "relying" on our grunt installation now is this:

```
3:02:05 PM: Executing user command: grunt
3:02:05 PM: grunt-cli: The grunt command line interface (v1.2.0)
3:02:05 PM: Fatal error: Unable to find local grunt.
3:02:05 PM: If you're seeing this message, grunt hasn't been installed locally to
3:02:05 PM: your project. For more information about installing and configuring grunt,
3:02:05 PM: please see the Getting Started guide:
3:02:05 PM: http://gruntjs.com/getting-started
```

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -512,7 +512,7 @@ install_missing_commands() {
   then
     if ! [ $(which grunt) ]
     then
-      npm install grunt-cli
+      npm install grunt-cli grunt
       export PATH=$(npm bin):$PATH
     fi
   fi


brycekahle:
This is on the [README](https://github.com/gruntjs/grunt-cli) for `grunt-cli`:
> The job of the grunt command is to load and run the version of Grunt you have installed locally to your project, irrespective of its version. Starting with Grunt v0.4, **you should never install Grunt itself globally**. For more information about why, please read this.

fool:
ok, so how do you propose we handle that error?  installing grunt-cli no
longer works under node 8.10.0

@futuregerald did some research to confirm this is a behavior change from
8.9.4 when using package.json to install just "grunt-cli" - but 8.10.0 is
the current LTS release.

brycekahle:
AFAIK you are supposed to install `grunt-cli` globally, but have `grunt` in your project's `devDependencies`.

fool:
ok.  It is my impression that this was not the case in the past, so this is
potentially breaking deploys for some users but we'll deal with them on a
case by case basis and re-escalate if there are more than a handful.

brycekahle:
Grunt v0.4 was released on Feb 18, 2013, so it has been a long time. We did not change anything about what global modules are installed in the image.

PR #167, opened 2018-03-24
Added Catalan, Basque and Galician

sukiletxe:
This PR adds some languages spoken in Spain, namely Catalan, Basque and Galician. (In my use case, I want Basque (but I can workaround it if this is not merged, and maybe I will regardless), but I thought I could add the rest, just in case somebody needs them in future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,8 @@ MAINTAINER Netlify
 #
 ################################################################################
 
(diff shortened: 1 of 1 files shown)


brycekahle:
@sukiletxe Thanks. This won't be available right away. I'll update this PR when it is.

Issue #151, opened 2018-02-23
Add support for Pipenv


jessebmiller:
If there is a Pipenv/Pipenv-lock file install dependencies with Pipenv.

If this is something you'd accept, I may be up for working on it.

brycekahle:
@jessebmiller Do you have any data on how popular `pipenv` is? I see python.org is now recommending it.

jessebmiller:
I don't have data on that, but with further research it looks like it's really just the `Pipfile` and `Pipfile.lock` formats that need support. `pipenv` is just the tool people are using until it's what `pip` supports it.

"Pipfile and its sister Pipfile.lock are a replacement for the existing standard pip's requirements.txt file."
- https://github.com/pypa/pipfile

jessebmiller:
Maybe this issue should be renamed to "Support Pipfile"

jayvdb:
Pipenv is a toml file, so quite a comfortable format for Netlify to inspect, and it includes the Python runtime

```
[requires]

python_version = "3.6"
```

It would be nice to use that instead of `runtime.txt`

jessebmiller:
That's wonderful!

PR #168, opened 2018-04-02
Support Pipfile

stlk:
As mentioned in #151 python community is heavily adopting dependency manager Pipenv. It's even recommended in official guide [Managing Application Dependencies](https://packaging.python.org/tutorials/managing-dependencies/#managing-dependencies). It combines virtualenv creation, runtime specification, dependencies and their locked versions into two files.

I added support for `Pipfile`. The script first checks for `runtime.txt` and then for `Pipfile`.

You can test the implementation on https://github.com/Liduska/pipenv-build-example, which is copy of repository I use with `requirements.txt` and `runtime.txt` removed. You can find original repository on https://github.com/Liduska/coffeeshop-netlify.

Thanks for the great product!

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -14,6 +14,10 @@ export GIMME_CGO_ENABLED=true
 export NVM_DIR="$HOME/.nvm"
 export RVM_DIR="$HOME/.rvm"
 
+# Pipenv configuration
+export PIPENV_RUNTIME=3.6
+export WORKON_HOME=$NETLIFY_CACHE_DIR/.pipenv
+
 YELLOW="\033[0;33m"
 NC="\033[0m" # No Color
 
@@ -159,6 +163,10 @@ install_dependencies() {
       echo "Please see https://github.com/netlify/build-image/#included-software for current versions"
       exit 1
     fi
+  elif [ -f Pipfile ]
+  then
+    echo "Found Pipfile installing Pipenv"
+    $HOME/python$PIPENV_RUNTIME/bin/pip install pipenv
   else
     source $HOME/python2.7/bin/activate
   fi
@@ -321,6 +329,24 @@ install_dependencies() {
       echo "Pip dependencies installed"
     else
       echo "Error installing pip dependencies"
+      exit 1
+    fi
+  elif [ -f Pipfile ]
+  then
+    echo "Installing dependencies from Pipfile"
+    if $HOME/python$PIPENV_RUNTIME/bin/pipenv install
+    then
+      echo "Pipenv dependencies installed"
+      if source $($HOME/python$PIPENV_RUNTIME/bin/pipenv --venv)/bin/activate
+      then
+        echo "Python version set to $(python -V)"
+      else
+        echo "Error activating Pipenv environment"
+        echo "Please see https://github.com/netlify/build-image/#included-software for current versions"
+        exit 1
+      fi
+    else
+      echo "Error installing Pipenv dependencies"
       exit 1
     fi
   fi
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ The specific patch versions included will depend on when the image was last buil
   * 6
   * 8 (default)
   * Any version that `nvm` can install.
-* Python - `runtime.txt`
+* Python - `runtime.txt` or `Pipfile`
   * 2.7 (default)
   * 3.4
   * 3.5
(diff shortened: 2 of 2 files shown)


brycekahle:
It looks like this is forcing python 3.6 if you want to use Pipenv. Is there some restriction on using that version of Python when using Pipenv?

stlk:
@brycekahle Ah sorry, I should have mentioned that. Python version set in `PIPENV_RUNTIME=3.6` variable is used to run Pipenv itself. Pipenv will then create environment based on version specified in `Pipfile`. It can be any version installed on the machine.

```
[requires]
python_version = "2.7"
```
You can see complete example at https://docs.pipenv.org/basics/#specifying-versions-of-python

brycekahle:
If there is no `python_version` in the `Pipfile`, would it create an environment based on `3.6`?

run-build-functions.sh
```diff
@@ -323,6 +332,24 @@ install_dependencies() {
       echo "Error installing pip dependencies"
       exit 1
     fi
+  elif [ -f Pipfile ]
+  then
+    echo "Installing dependencies from Pipfile"
+    if $HOME/python$PIPENV_RUNTIME/bin/pipenv install
+    then
+      echo "Pipenv dependencies installed"
+      if source $($HOME/python$PIPENV_RUNTIME/bin/pipenv --venv)/bin/activate
```
jayvdb: fwiw, doing this instead of `pipenv shell` and `pipenv run` means `.env` loading is not supported, which could break some `Pipfile` if it is expecting variables from `.env` to be present.

https://github.com/pypa/pipenv/blob/master/docs/advanced.rst#-automatic-loading-of-env
https://github.com/pypa/pipenv/blob/master/docs/advanced.rst#-support-for-environment-variables

Note that it isnt even possible to use `.env` vars during `pipenv install`.  see https://github.com/pypa/pipenv/issues/1906#issuecomment-378635177

As a result, probably a good idea to set `PIPENV_DONT_LOAD_ENV` and document that .env isnt used by Netlify.

run-build-functions.sh
```diff
@@ -323,6 +332,24 @@ install_dependencies() {
...
+      then
+        echo "Python version set to $(python -V)"
+      else
+        echo "Error activating Pipenv environment"
+        echo "Please see https://github.com/netlify/build-image/#included-software for current versions"
```
jayvdb: this isnt particularly useful here, as they would go to that page to learn that "Latest version." is the current version, and they would find the current version was actually in the log a few lines above this. ;-)

run-build-functions.sh
```diff
@@ -323,6 +332,24 @@ install_dependencies() {
       echo "Error installing pip dependencies"
       exit 1
     fi
+  elif [ -f Pipfile ]
+  then
+    echo "Installing dependencies from Pipfile"
+    if $HOME/python$PIPENV_RUNTIME/bin/pipenv install
```
jayvdb: might be useful to set `PIPENV_VENV_IN_PROJECT` as multiple venvs are not needed on Netlify builds.  setting that might improve performance a little.

run-build-functions.sh
stlk: Well, I thought this refers to list of python versions user can use.

run-build-functions.sh
stlk: Thanks for the feedback! It seems to me that ENV variables are meant to be handled some other way on production. E.g. using https://www.netlify.com/docs/continuous-deployment/#build-environment-variables.

Based on discussion in https://github.com/pypa/pipenv/issues/773#issuecomment-333246959 `pipenv shell` is meant only for development and using `source .../bin/activate` is OK for production.

run-build-functions.sh
stlk: Good point, thanks! 🥇

jayvdb:
ping @brycekahle

brycekahle:
@jayvdb Sorry for the delay. I need some time to look at this more in depth.

run-build-functions.sh
```diff
@@ -14,6 +14,11 @@ export GIMME_CGO_ENABLED=true
 export NVM_DIR="$HOME/.nvm"
 export RVM_DIR="$HOME/.rvm"
 
+# Pipenv configuration
+export PIPENV_RUNTIME=3.6
```
brycekahle: I think I would prefer this to be `2.7` if possible, since that is our default python version at the moment.

run-build-functions.sh
brycekahle: I think we should move this (and the `export` lines from above) to the `Dockerfile` if possible. I'd rather `pipenv` be installed already, rather than done every build on-demand.

stlk:
@brycekahle Thanks for the feedback! I made the changes you requested.

PR #170, opened 2018-05-04
Install ruby 2.5.0 with option '--disable-binary'

unasuke:
Install ruby 2.5.0 to ubuntu 14.04 with rvm have an issue reported at https://github.com/rvm/rvm/issues/4291 .
So, add the option "--disable-binary" to install ruby 2.5.0 to avoid the issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -211,10 +211,12 @@ RUN gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys D39DC0E3 && \
 
 ENV PATH /usr/local/rvm/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
+# about the option of the install ruby 2.5.0, see https://github.com/rvm/rvm/issues/4291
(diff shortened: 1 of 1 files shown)


migueldemoura:
Why install `2.5.0` by default when `2.5.1` has been released and includes security patches?

unasuke:
@migueldemoura  If you want to use ruby 2.5.1, set `RUBY_VESION` to `2.5.1`. That's all.
But a user that want to use ruby 2.5.0, it fails to install ruby 2.5.0 with set the env to `2.5.0`.
Because install ruby 2.5.0 into ubuntu 14.04 by rvm needs the option `--diable-binary`.

joshRpowell:
ping @brycekahle

migueldemoura:
> @migueldemoura If you want to use ruby 2.5.1, set RUBY_VESION to 2.5.1. That's all.

@unasuke, that's the problem I've explained on the other thread. The currently deplyed image doesn't support `2.5.1`. It crashes when executing `bundler`.

brycekahle:
The upstream issue has been identified, so we are going to wait until they release the binary fixes.

Let me know if you have any questions.

PR #171, opened 2018-05-08
Install Google Bazel build tool

mboes:
Google [Bazel] is a general-purpose polyglot build tool. Some projects frequently use it as a replacement for make or other build tools for generating static content, e.g. for generating API documentation from sources. It has rules available for building typescript code, Sass/CSS and more.

The build image already pulls in other build tools like CMake. Google Bazel is both general-purpose and fairly lightweight given that the main dependencies (Java) are already part of the image. In fact adding the `bazel` package induces no new dependencies apart from maybe `bash-completion`.

[Bazel]: https://bazel.build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,8 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     update-locale en_US.UTF-8 && \
     apt-key adv --fetch-keys https://packagecloud.io/github/git-lfs/gpgkey && \
     apt-add-repository -y -s 'deb https://packagecloud.io/github/git-lfs/ubuntu/ trusty main' && \
+    echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list && \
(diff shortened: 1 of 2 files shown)


brycekahle:
@mboes Can you describe your specific use case?

mboes:
One concrete use case is this: https://haskell.build. This site hosts API documentation. The documentation is generated from source code. Currently it does that within a Docker container provided to Heroku. It would be simpler to use Netlify for this, since the resulting content is all static. For this common use case, i.e. generating sites at least partially from source code artifacts (such as API documentation), I wouldn't know how to do it in Netlify instead of Heroku if I can't simply ask my build tool to perform static website generation (via Doxygen, Skydoc, etc).

brycekahle:
@mboes According to this [documentation](https://docs.bazel.build/versions/master/install-ubuntu.html), you should be able to do a user binary installation of bazel. We have all the dependency packages as part of the image already.

brycekahle:
@mboes Can you give that a try and let me know if it works?

mboes:
I tried it that way. It does work, at the cost of an extra 2 minutes on build time. I can live with that.

PR #176, opened 2018-06-22
Add standard directory tree generator

techtonik:
tree -H '.' -L 1 > index.html


brycekahle:
@techtonik Can you describe your use case of `tree`?

Issue #141, opened 2018-02-18
Support specifying which languages to install/skip


edmorley:
Hi!

We have an app that is comprised of a webpack-built SPA and a django-rest-framework powered backend - with both parts existing in the same repository (monorepos have [many advantages](https://danluu.com/monorepo/)):
https://github.com/mozilla/treeherder

We would only want the SPA to be built by Netlify (so only want the nodejs env setup + `yarn install` steps to be run), however `run-build-functions.sh` also finds `runtime.txt` and `requirements.txt`, resulting in unwanted (and failing) Python setup tasks being run:

```
7:13:05 PM: Installing dependencies
7:13:05 PM: /usr/local/bin/run-build-functions.sh: line 150: /opt/buildhome/pythonpython-2.7.14/bin/activate: No such file or directory
7:13:05 PM: Error setting python version from runtime.txt
...
7:13:08 PM: Installing pip dependencies
...
7:13:12 PM: Collecting mysqlclient==1.3.12 (from -r requirements/common.txt (line 74))
7:13:12 PM:   Downloading mysqlclient-1.3.12.tar.gz (89kB)
7:13:12 PM:     Complete output from command python setup.py egg_info:
...
7:13:12 PM:     EnvironmentError: mysql_config not found
7:13:12 PM:     ----------------------------------------
7:13:13 PM: Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-PcfhfO/mysqlclient/
7:13:13 PM: Error installing pip dependencies
7:13:13 PM: Error running command: Build script returned non-zero exit code: 1
7:13:13 PM: Failing build: Failed to build site
```

The "install specific Python version" step is giving a warning, since our `runtime.txt` is for use by [heroku-buildpack-python](https://github.com/heroku/heroku-buildpack-python) and so uses their `python-N.N.N` format rather than that used by Netlify (ie has the additional `python-` prefix).

The pip install step is failing since the `mysqlclient` package requires that the C `libmysqlclient-dev` be installed.

However even if these two steps succeeded, they would only be unnecessarily increasing our build times, since our SPA has no need for the Python environment.

As such we would like a way to prevent them from being run.

Perhaps a new `language` option, like:
```toml
# netlify.toml
[build]
  language = node
  publish = "dist"
  command = "yarn build"
```

...which would skip all steps other than those relating to node/js dependencies.

Or else a `LANGUAGES` environment variable would be fine too.

Many thanks! :-)

brycekahle:
[...]

PR #187, opened 2018-08-09
Make nix available in the container

wereHamster:
This makes [nix](https://nixos.org/nix/) available in the build image. It will allow users to use `nix-shell` or `nix-build` to build the websites.

I tested the image by building it and then executing `nix-shell` inside it:

```
docker run -ti <image> \
  nix-shell -p nodejs-10_x \
    --command "node -e 'console.log(process.version)'"
```

(here nix-shell is instructed to make node.js v10 available and then run `"node -e 'console.log(process.version)'"`).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -405,9 +405,10 @@ RUN mkdir -m 0755 /nix && chown buildbot /nix
 USER buildbot
 WORKDIR /tmp
 RUN curl -o install-nix https://nixos.org/nix/install && USER=buildbot sh ./install-nix && rm install-nix
+RUN mkdir -p /opt/buildhome/.nix/bin
 
-ENV PATH=/nix/var/nix/profiles/per-user/$USER/profile/bin:/nix/var/nix/profiles/per-user/$USER/profile/sbin:$PATH
-ENV NIX_PATH=/nix/var/nix/profiles/per-user/$USER/channels
+ENV PATH=/opt/buildhome/.nix/bin:$PATH
+ENV NIX_PATH=/nix/var/nix/profiles/per-user/buildbot/channels
 ENV NIX_SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt
 
 # Cleanup
--- a/Dockerfile
+++ b/Dockerfile
@@ -394,6 +394,22 @@ ENV GIMME_GO_VERSION "1.10"
 ENV GIMME_ENV_PREFIX "/opt/buildhome/.gimme/env"
 RUN gimme
 
+################################################################################
+#
+# Nix
+#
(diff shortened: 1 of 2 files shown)


wereHamster:
Don't pay too close attention to the fact that I used node in the test, nix actually provides thousands of packages. To test the build-image I could also have used `docker run -ti <image> nix --version`.

wereHamster:
Dependencies that are specified in nix files take precedence over whatever is installed by the system. Furthermore, `nix-shell` has the `--pure` flag to explicitly ignore anything that is installed globally. I can't find that same flag in the `nix-build` documentation, but I'd expect it to behave like `nix-shell --pure` by default.

brycekahle:
A major downside of this, is having to install those packages for **every** build. Our current setup goes through a lot of effort to prevent that.

grossbart:
I'm not sure I understand the Netlify architecture well enough, but I would assume that the Docker image is built only once per project? If that is the case then Nix will take care of your concern: it builds all dependencies only once and stores them in the Nix store, each subsequent start of a Nix shell will then use the already built dependencies, which takes no time at all.

brycekahle:
@grossbart That isn't how our CI system works. We have a global base docker image, then we save/restore caches of certain directories from blob storage during each build.

wereHamster:
Nix makes it possible to redirect all artefacts (dependencies and intermediate build products) to a separate directory (see my last commit). So all good… except…

…there doesn't seem to be a way to skip the pre-processing and dependency installation that the build script does on its own. I tested it with one of my projects where nix reliably works, but it fails inside the netlify build container because the build script tries to (automagically) install dependencies using the wrong node/yarn versions. I was able to work around this by setting NODE_VERSION and YARN_VERSION, but that kindof defeats the purpose of using nix (one of the reasons to use nix was so that we don't have to specify the dependencies in multiple places).

brycekahle:
@wereHamster We are tracking the ability to skip certain steps in #141

bcomnes:
Supporting a local nix or homebrew would be very cool indeed.  I'm taking over for bryce on the buildbot work.  So, open to it!  But I don't have any kind of timeline yet.

bcomnes:
I changed the base to `xenial`, since if we do this, we'll do it on that branch.

thalesmg:
I was about to request Nix support too. I'm building some simple pages using it, and I'm currently only able to deploy that using Travis' `dpl` tool. But I miss using Netlify's Pull Request previews and integrations...

domenkozar:
Is there a sensible path forward here?

wereHamster:
Someone else can pick this up. I no longer plan working on this.

PR #188, opened 2018-08-13
Add erlang and elixir

vorce:
Install erlang and [Elixir](https://elixir-lang.org/) in the build image. This will allow continuous deployments to use elixir (and erlang) to build websites.

I tested the image by building it and then executing `elixir --version` inside it:

```
❯ docker run -ti <image> elixir --version
Erlang/OTP 21 [erts-10.0] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.7.0 (compiled with Erlang/OTP 20)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,9 +43,9 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         bzr \
         cmake \
         curl \
+        elixir \
         emacs25-nox \
         esl-erlang \
-        elixir \
         expect \
         fontconfig \
         fontconfig-config \
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     echo 'LANGUAGE="en_US:en"' >> /etc/default/locale && \
     locale-gen en_US.UTF-8 && \
     update-locale en_US.UTF-8 && \
+    apt-key adv --fetch-keys https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc && \
     apt-key adv --fetch-keys https://packagecloud.io/github/git-lfs/gpgkey && \
     apt-add-repository -y -s 'deb https://packagecloud.io/github/git-lfs/ubuntu/ trusty main' && \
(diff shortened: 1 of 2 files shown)


vorce:
Hey @brycekahle, any interest in support for Elixir and thoughts on the PR? Thanks!

Dockerfile
```diff
@@ -394,6 +394,19 @@ ENV GIMME_GO_VERSION "1.10"
...
+################################################################################
+#
+# Erlang+Elixir
+#
+################################################################################
+USER root
+# Erlang solutions repository
+RUN wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb && \
+    dpkg -i erlang-solutions_1.0_all.deb
+RUN apt-get update
```
brycekahle: Is it possible to add this repository using `add-apt-repository`? If so, it would be better to include these installs in the giant `apt` command at the start of the `Dockerfile`. That way it only has to run `apt-get update` once.

brycekahle: It looks like the official erlang instructions under "Alternatively: adding the repository entry manually" apply. I think this would do it:

```
apt-key adv --fetch-keys https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc && \
apt-add-repository -y 'deb https://packages.erlang-solutions.com/ubuntu trusty contrib' && \
```
and lower down in the list of packages add `esl-erlang` and `elixir`.

vorce: Cool! I will try this and update the PR once I got something working.

vorce:
@brycekahle updated PR with `apt-add-repository`, it's a bit less clear where elixir is installed now though, but maybe not really a problem.

Dockerfile
```diff
@@ -42,6 +44,8 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         cmake \
         curl \
         emacs25-nox \
+        esl-erlang \
```
brycekahle: Could you put these in alphabetic order? Thanks!

vorce: Doh. Done

PR #191, opened 2018-08-24
document versions of some important software

slang800:
I've been running into a lot of version-related issues lately, so I think this is useful documentation to include.


brycekahle:
@slang800 Thanks!

Issue #183, opened 2018-07-10
install necessary packages for using a CGO-extension compiled version of Hugo


fool:
When attempting to run a new hugo build, it seems we are missing some dependencies:

```
6:20:18 PM: Executing user command: ./hugo --config paceup.com.toml,config.toml
6:20:18 PM: ./hugo: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./hugo)
6:20:18 PM: ./hugo: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./hugo)
```

We need to install whatever is needed here :)

fool:
Should notify these customers in our helpdesk:

-  https://app.intercom.io/a/apps/q245f50x/inbox/inbox/unassigned/conversations/17497593067

regisphilibert:
Untill then it’s hard to mention Netlify (let alone use it) alongside Hugo Pipes... too bad because speeding up deployement is among the big pros 😔

DirtyF:
@regisphilibert https://hugoscss.netlify.com/ :)

regisphilibert:
@DirtyF this repo commits its resources directory. It’s really not ideal if the assets can be edited by an editor.

regisphilibert:
Hi @oncletom, let's take this here: https://discourse.gohugo.io/t/netlify-and-hugo-pipes/13145

jhabdas:
If Netlify's CI supports Docker the following Dockerfile should get you up and running. Builds Hugo from source. Set CGO env var from file or override via CLI:

https://git.habd.as/comfusion/after-dark/src/branch/master/docker/hugo/Dockerfile

brycekahle:
Appears it was intended to be a dynamically linked binary: https://github.com/gohugoio/hugo/issues/4985

jensamunch:
this doesn't work still for hugo 0.46

what is the best thread for tracking this issue and netlify using the extended version?

brycekahle:
@jensamunch This issue is tracking it.

fool:
hugo 0.46 works well for me - just can't use the extended version, right @jensamunch ?

oncletom:
[...]

Issue #182, opened 2018-07-10
Support extended Hugo version


DirtyF:
Duplicate of: https://github.com/netlify/binrc/issues/16
Opening an issue here as @futuregerald kindly suggested.

**- What is the current behavior?**

From version 0.43 Hugo comes in two flavours, extended version allows native Sass compilation.
Netlify currently does not support extended version:

```
2:28:30 PM: ERROR 2018/07/10 12:28:30 error: failed to transform resource: failed to transform “scss/main.scss” (text/x-scss): this feature is not available in your current Hugo version
```

**Edit: For now, the recommended workflow is to commit your `/resources` folder after building locally with `hugo`.**

**- What is the expected behavior?**

Either download the extended binary by default from now on or allow to specify something like 0.43/extended in netlify.toml.

_extended is added to the release path before the version number:

https://github.com/gohugoio/hugo/releases/download/v0.43/hugo_0.43_Linux-64bit.tar.gz
https://github.com/gohugoio/hugo/releases/download/v0.43/hugo_extended_0.43_Linux-64bit.tar.gz

jensamunch:
i tried manually including binary in the repo - as a quick-fix - doesn't work:

```
6:20:18 PM: Executing user command: ./hugo --config paceup.com.toml,config.toml
6:20:18 PM: ./hugo: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./hugo)
6:20:18 PM: ./hugo: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./hugo)
```

fool:
We'll track that separate issue here, @jensamunch : https://github.com/netlify/build-image/issues/183

baublet:
To those looking for a workaround until this is fixed:

Run `hugo` locally with the extended version, then check in your `resources/_gen/assets` directory to the repository so that Hugo skips trying to generate the assets.

jensamunch:
It would be really great to have a bit more of an explanation from the Netlify team on timeline and plan.

brycekahle:
@jensamunch Unfortunately, the solution involves upgrading the version of Ubuntu we use in the base image. That is a larger user impact than we initial hoped for. In order to ensure a smooth transition for all users, we need to do some work around communicating the changes ahead of time. Once we have a firm timeline, we will communicate that publicly.

igk1972:
[...]

PR #197, opened 2018-09-07
Add Hugo extended (>0.43)

igk1972:
Uncompress necessary files from one deb-package for using a CGO-extension compiled version of Hugo (>0.43).
See https://github.com/netlify/build-image/issues/183 and https://github.com/netlify/build-image/issues/182 

And set expand alias with LD_LIBRARY_PATH - only for hugo extended.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -323,6 +323,7 @@ RUN binrc install spf13/hugo 0.17 -c /opt/buildhome/.binrc | xargs -n 1 -I{} ln 
     binrc install spf13/hugo 0.18 -c /opt/buildhome/.binrc | xargs -n 1 -I{} ln -s {} /usr/local/bin/hugo_0.18 && \
     binrc install spf13/hugo 0.19 -c /opt/buildhome/.binrc | xargs -n 1 -I{} ln -s {} /usr/local/bin/hugo_0.19 && \
     binrc install spf13/hugo 0.20 -c /opt/buildhome/.binrc | xargs -n 1 -I{} ln -s {} /usr/local/bin/hugo_0.20 && \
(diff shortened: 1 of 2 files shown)


brycekahle:
@igk1972 We appreciate the PR, but I don't think we are comfortable installing packages designed for Ubuntu 16 on Ubuntu 14. The risk for negative side-effects is too high.

igk1972:
> @igk1972 We appreciate the PR, but I don't think we are comfortable installing packages designed for Ubuntu 16 on Ubuntu 14. The risk for negative side-effects is too high.

But the package is NOT installed on the system. The library is loaded only separately and it is only for hugo-0.43

brycekahle:
> But the package is _not_ installed on the system

I see that now. Let me discuss this with some colleagues.

bluestrike2:
> > But the package is _not_ installed on the system
> 
> I see that now. Let me discuss this with some colleagues.

@brycekahle 

Any movement on the subject, by chance?

brycekahle:
@bluestrike2 I'm in the process of leaving Netlify. Let's get @rybit involved here. He can keep you up to date on discussions.

bcomnes:
Thank you @igk1972 for this, sorry its taken so long, we are considering going with this as a short term solution now.

bcomnes:
After testing this, this will be the interim solution we are using until we can make larger changes to the build-bot.

bcomnes:
An issue with this: Makefile's that reference hugo won't get the alias.  Brainstorming a fix.

bcomnes:
Since this is going to cause more issues for the current situation than without it, I'm going to partially revert some of this.  We'll still keep the shared lib additions, but we won't install extended by default or alias the LD env vars in, since they aren't reliable in different shell contexts.  Also, we are going to expedite making a Ubuntu 16 image available.  Hang tight for more details.

PR #202, opened 2018-10-01
Report lingering processes to the user

sgirones:
Netlify builds will wait until all the processes finish. If they don't finish, the build will eventually timeout and will be marked as failed.

This code will detect and report any lingering process after the main build command finishes so the user can make sure they are terminated properly.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -623,3 +623,32 @@ set_go_import_path() {
     cd $importPath
   fi
 }
+
+find_running_procs() {
+  ps aux | grep -v [p]s | grep -v [g]rep | grep -v [b]ash
+}
+
+report_lingering_procs() {
+  procs=$(find_running_procs)
+  nprocs=$(expr $(echo "$procs" | wc -l) - 1)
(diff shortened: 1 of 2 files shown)


rybit (approved):

bcomnes:
I'm going to rebase and see if that fixes the builds.

bcomnes:
Oh wait, there is no place to rebase too 🤔

bcomnes:
Seems like there are some dependency issues to resolve first.  Fixing those, then we'll rebase this.

bcomnes:
Rebased, and rebuilding.

bcomnes:
screen shot 2018-11-14 at 5 34 44 pm

Looks like its working!  Not sure what that huge string is though.  @sgirones any ideas?

sgirones:
argh, that's the buildbot process itself. We should exclude it from the list of processes

sgirones:
Nevermind, I saw your new PR

bcomnes:
Was a null test with no lingering process.  I'll perform a test with a lingering process, but after I filter out the build bot itself.

PR #212, opened 2018-11-05
Update ruby gems bundle caching

kitop:
There was a two-part typo for using the cached path when installing ruby gems.

In line 35 cache directory is defined as `$NETLIFY_CACHE_DIR/.bundle`, but in `--path` argument to `bundle install` it as passed as `$NETLIFY_CACHE_DIR/bundle` without the dot.

Additionally, because it's doing `restore_cwd_cache ".bundle"` before (and `cache_cwd_directory ".bundle"` at the end), that directory is moved to `$PWD` (and back after), so `$NETLIFY_CACHE_DIR/.bundle` is always empty when running `bundle install`, causing gems to be installed again.

This commit changes it so the `--path` argument to `bundle install` is simply `.bundle` in the current working directory so the existing caching mechanisms can be kept in place and bundler actually uses the gem cache.
There's no need to change the `--binstubs` argument because bundle will re-generate those even if gems are already installed (cached in this case).

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -312,7 +312,7 @@ install_dependencies() {
     if install_deps Gemfile.lock $RUBY_VERSION $NETLIFY_CACHE_DIR/gemfile-sha || [ ! -d .bundle ]
     then
       echo "Installing gem bundle"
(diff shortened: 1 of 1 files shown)


bcomnes:
Spoke with @kitop today about this, looks like caching is working as expected, so maybe this is not needed?  Perhaps it was fixed elsewhere, anyway, we'll hold this open until we know its not needed.

bcomnes:
Closing but we can reopen if needed

PR #225, opened 2018-11-27
Add sha and tag docker tags

bcomnes:
In order to facilitate more fine-grained buildbot builds, we need a more fine-grained docker tagging strategy.  

This PR does the following: 

- Add COMMIT tags to our docker builds
- Adds COMMIT-squash tags to our docker builds
- Add TAG and TAG-squash docker tags when tagged

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,8 +16,8 @@ pipeline {
         branch 'master'
       }
       steps {
-        sh "docker build --build-arg NF_IMAGE_VERSION=${env.GIT_COMMIT} -t netlify/build:latest ."
-        sh "docker build --build-arg NF_IMAGE_VERSION=${env.GIT_COMMIT} --squash -t netlify/build:squash ."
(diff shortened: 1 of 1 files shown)


sgirones:
The problem I see with using only Jenkinsfile is that there's no way we can test it locally. I would rather have all those steps in a script that you can run and debug.

But just my two cents!

bcomnes:
@sgirones agreed that the ci flow should be trivially runably locally.  It ended up being a bit simpler than this in the end (once I figured out the behavior of the git ENV vars, and a bunch of time out of my day).  https://github.com/netlify/build-image/blob/v2.6.7/Jenkinsfile

I think I saw buildbot's cloud builder come with a similar tagging behavior by default, so maybe moving over to that would simplify things even furthur.  

If you have a preferred CI pattern for docker image baking, in Jenkins or wherever, I would be happy to consider moving over to it.  I'm not attached to this one beyond its what we are running now.

bcomnes:
@sgirones I'm thinking maybe I didn't understand the original suggestion in the above reply.  

Do you mean move all shell scripts into their own file similar to https://github.com/netlify/buildbot/blob/master/script/docker-build.sh ? 

Yes! Agreed, if thats the case.  I'll clean it up at some point.

sgirones:
yep! sorry I didn't get back to you.

PR #244, opened 2018-12-21
Add Hugo 0.52 to dev image

bcomnes:


bcomnes:
I tested this back when we tried to do this in the production image.  This portion worked great, so lets go with it.

PR #246, opened 2019-01-03
Fix Bundler version to 1.17.3 to prevent 2.0 being installed in ruby versions < 2.6

### LINKED ISSUE #245: Builds failing with a Ruby error.
Recent builds are failing with the following error:

```
5:00:32 AM: ERROR:  Error installing bundler:
5:00:32 AM: 	bundler requires RubyGems version >= 3.0.0. Try 'gem update --system' to update RubyGems itself.
5:00:32 AM: Error installing bundler
```

This also affects builds that don't use Ruby at all.

Possible fix is to update default Ruby version to 2.6.0 here: https://github.com/netlify/build-image/blob/master/run-build-functions.sh#L282

kitop:
[Bundler 2.0](https://rubygems.org/gems/bundler/versions/2.0.0) was released today (January 3rd, 2019). It requires rubygems 3.0.0 and drops support for ruby < 2.3. This is causing builds using custom ruby < 2.6 to fail.

In this change, we set the bundler version to 1.17.3—the current last one before 2.0.0—, so it keeps working with existing ruby setups.

Closes #245.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -237,9 +237,9 @@ RUN gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys D39DC0E3 && \
 ENV PATH /usr/local/rvm/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
 RUN /bin/bash -c "source ~/.rvm/scripts/rvm && \
-                  rvm install 2.2.9 && rvm use 2.2.9 && gem install bundler && \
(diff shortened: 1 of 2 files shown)


bcomnes:
Thank you @kitop I'll get this in.

bcomnes:
We probably dont need this now https://bundler.io/blog/2019/01/04/an-update-on-the-bundler-2-release.html

If this is still an issue, please let me know what it is.

PR #272, opened 2019-03-08
Add TeX Live

vermiculus:
These patches install the most recent TeX Live on the image.  Having a TeX system allows building documents from source before making them available as PDFs on the site.  (I've been doing this successfully for some time to build reference documentation by using a custom build script.)  Folks who maintain academic sites would also appreciate the availability of TeX by default to build general papers.

I've split the change into two commits to allow for flexibility in how this is implemented.  The most straight-forward way to do this is given by the sum of the two commits: create a TeX Live install profile file (AFAIK required to run non-interactively) and `ADD` and load that profile in the Dockerfile.  Alternatively, we can just use the first commit: generate this file on-demand by using `printf ... >file`.

The only thing I'm wholly unsure about is the installation directory, but using `/tmp` has not caused problems in the few years I've been building documents on Netlify.

Tested the build with `docker run <some-sort-of-unique-id> xelatex --version` and received the right version (currently TeX Live 2018).

This is also my first exposure to Docker and the `Dockerfile` format, so let me know if there's something I should change!  :smile:

---

I should add: you might ask 'Why not just use one of the existing apt packages for TeX Live?' Given the difficulty of packaging, these packages are usually several *years* behind recent development. It's incredibly frustrating to know that an issue was fixed two years ago but your installation is older and difficult to adjust :-)


bcomnes:
@vermiculus Very cool!  I would love to see this get added if we can manage the extra size of TeX in the image.

vermiculus:
Thanks! If size becomes a concern, we can alter the installation scheme in `texlive.profile` to a more appropriate value.

bcomnes:
I'm going to change the base of this to trusty since we are going to remove the master branch in favor of the named branches.

bcomnes:
Also, I'll need to cherry pick this off of the xenial branch, since we are going to freeze trusty soon.

vermiculus:
Ah, branches per OS? Makes sense in this context.

Resolved the merge conflict with the README.

vermiculus:
Just a ping – is there any information I can provide to help get this merged? 😄

bcomnes:
Sorry this has taken so long.  My main concern is size of this addition and I haven't had enough time to characterize it in order to make a decision one way or another.

cs:
@bcomnes I don't know about the original author's requirements, but I would be happy with `groff` too.

I need LaTeX only to typeset mathematical formulae, but, of course, I understand your desire to keep the build images as small as possible. Installing `groff` is much easier because the standard Ubuntu package is sufficient: `apt-get install groff`. This would increase the build image size by ~30 MB to a total of ~3.88 GB. I'd be happy to open a PR if this sounds reasonable. Please let me know.

vermiculus:
I'm also sure the TeX Live installation profile can be adjusted to whatever size requirements you may have – [down to 76 MB](https://tex.stackexchange.com/a/397181/17423). (Honestly, even just removing all `texdoc` stuff from the image would save a big chunk of space.)

---

As for my requirements, I'm providing a full-on document with text/images. Little/no math, in fact, but formatting requirements are beyond the capabilities of tools like nroff/groff.

@cs, if you only need LaTeX to typeset maths but are otherwise happy with groff, you should look into [eqn](https://en.wikipedia.org/wiki/Eqn_(software)) It's obviously not as powerful as TeX, but it may be sufficient for your needs.

vermiculus:
Turns out that I'd already removed all the `texdoc` stuff, but I've added a commit that changes to using the 'basic' scheme.  I'm unable to see what failed in CI though (I'm getting a 401-unauthorized). Do I need to rebase? (If so, what should be the new base?)

cs:
@vermiculus Nice! As I said yesterday, LaTeX and `groff` are both sufficient for my purpose. If we get `groff`, I'm planning to invoke `eqn2graph` from my static site generator. Otherwise, if we get LaTeX, I will probably use `pdflatex`.

In the end, I meed my formulas rendered to PNG. Right now, I think this is not possible, because the build image has literally no way to convert PDF => PNG. ImageMagick doesn't work because PS/PDF support is disabled for security reasons. Anyways, that's a problem for another day.

mraerino:
oh, this hasn't been touched for a long time 🙈 
I hope we can land #411 soon, so you could maybe just use this?
https://formulae.brew.sh/formula-linux/tectonic

vermiculus:
Interesting idea. Would it cache the image after installing tooling? That's the build slowdown.

mraerino:
yeah, homebrew keeps any previously installed stuff in the netlify cache, so install time on subsequent requests should be minimal.

you can actually try it out here: https://build-image-beta.netlify.app/
warning: since that image is not preloaded on the build boxes it will take some time to pull down when the build starts, so maybe try on a test site

vermiculus:
#411 should work :-) Out of curiosity, does it work with custom taps? (Truly just a curiosity; I believe [this package](https://formulae.brew.sh/formula-linux/texlive) would be more than adequate.)

Closing under the assumption that TeX Live can be installed simply and cached.

9600:
> #411 should work :-) 

Doesn't appear to for me. Adding texlive to Brewfile.netlify resulted in:

> 9:35:48 AM: Error: texlive: no bottle available!
> 9:35:48 AM: You can try to install from source with:
> 9:35:48 AM:   brew install --build-from-source texlive
> 9:35:48 AM: Please note building from source is unsupported. You will encounter build
> 9:35:48 AM: failures with some formulae. If you experience any issues please create pull
> 9:35:48 AM: requests instead of asking for help on Homebrew's GitHub, Twitter or any other
> official channels.

Confirmed that [no bottle is available](https://formulae.brew.sh/formula-linux/texlive) and building from source would seem like a bad idea.

Could we reopen this issue?

EDIT: just spotted that Tectonic was being suggested for Homebrew install instead of TeX Live. Looking into whether this can be used the same way with Sphinx, rendering both PDF and HTML output.

vermiculus:
@9600 Note that this wasn't an issue, it was a pull request -- if you're looking for support, it's probably best to open a new issue. Do mention this PR from there, though -- perhaps these changes may yet be valuable for you.

PR #274, opened 2019-03-11
Suffix binrc cache with version number

bcomnes:
This is to prevent caching overlap with other versions of binrc

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -436,7 +436,7 @@ install_dependencies() {
   if [ -n "$HUGO_VERSION" ]
   then
     echo "Installing Hugo $HUGO_VERSION"
-    hugoOut=$(binrc install -c $NETLIFY_CACHE_DIR/.binrc hugo)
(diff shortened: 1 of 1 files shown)


rybit:
could you explain how you ran into this? Symptoms etc so we know if it is a problem.

bcomnes:
I just explained on my 1:1, but for whoever may come across this:

Here is what we saw:

- The version of binrc differs between trusty and xenial.  The version on xenial downloads hugo extended, trusty downloads non-extended. 
- A site that was on trusty and had a seeded cache is switched over to xenial.  When binrc tries to install hugo, it looks in the cache for the version it wants, and uses it if it finds it instead of downloading again.  
- This resulted in a xenial image using a non-extended version of hugo that it would restore from the cache. 
- This behavior was observed, even after you clear the cache.  All restored caches would still have the old version of hugo.  It was almost as if the cache wasn't actually getting cleared, and new cache saves would include old versions of stuff that were added from the trusty image.  
- This prefix trick basically invalidates the binrc cache by adding a version number suffix to the folder it uses, thus invalidating the cache.

PR #247, opened 2019-01-07
Add Clojure CLI tools

joonashak:
Clojure's own CLI tools for running Clojure programs, REPLs, compiling, etc.

[https://clojure.org/guides/deps_and_cli](https://clojure.org/guides/deps_and_cli)

PR #279, opened 2019-03-13
Add Clojure CLI tools

bcomnes:
Rebase of https://github.com/netlify/build-image/pull/247


bcomnes:
Ok, this didn't work unfortunately, and I don't have time to look into clojure specifically so closing.  Hopefully we can land https://github.com/netlify/build-image/pull/187 or homebrew-linux and people can just self serve these tools.

PR #282, opened 2019-03-18
Add Go version to included software list

verythorough:
This will need to be added to the trusty branch, too. Want to just cherry-pick it over after merging?


bcomnes (approved):
Thank you, good catch

bcomnes:
Cherry-picking next

PR #288, opened 2019-03-21
Remove tmp files

bcomnes:
This clears out tmp files


rybit:
was this a problem?

bcomnes:
I'm hoping this results in a smaller squashed build image.

rybit:
If you launch the image, do you have files there?

bcomnes:
Yeah, about 450-500mb of files, looks like leftovers from various install processes.

rybit:
nice catch

PR #297, opened 2019-04-11
Added Netlify environment variable.

bkuhlmann:
## Overview

For downstream authors of linting/build tools that run in various CI
environments, it's important to be able to detect which environment the
build tool is being run in. This is important because of how Netlify
configures its environment, environment variables, dependencies, etc.
as they different from other CI environments like Travis CI or Circle
CI for example.

To illustrate further, when running on Circle CI or Travis CI, the
following environment variables are always present:

    CIRCLECI=true
    TRAVIS=true

The change being committed here adds Netlify support in the same manner
as found in other CI environments:

    NETLIFY=true

This provides consistency of behavior across the various CI
environments so authors, like me, can correctly detect which build
environment my tool is being run in and adjust accordingly.

This is of greatest benefit for tools like [Git Cop](https://is.gd/UyxZGq) that needs to calculate commits on the feature branch for linting purposes. If this pull request should be accepted, I'll be able to update my [documentation](https://github.com/bkuhlmann/git-cop#continuous-integration-ci) accordingly to show I can support Netlify properly.

## Details

This will help resolve a [workaround](https://github.com/bkuhlmann/git-cop/pull/45) I have in place currently based off of @fool's support feedback where I'm currently using the `DEPLOY_URL` to detect if the `netlify` string is present.

## Notes

This work is based off of feedback I got from @overlordofmu.


overlordofmu:
I also agree with logic for this change and would love to see this PR merged.

imorente:
+1 

When I added the [Netlify CI integration to Danger JS](https://github.com/danger/danger-js/pull/760/files#diff-336748283d435828dac653093eb577e5R21) I relied on an undocumented env variable –`NETLIFY_BUILD_BASE`. Would have been great to have something more reliable like `NETLIFY=true`

bcomnes:
We can probably do this.

bkuhlmann:
Hello. Wanted to check back in and see if there is any further traction on this? Having this land would allow me to clean up some code on my side. :wink:

bcomnes:
Yeah this is an easy add, and we don't have an equivalent (although we have other unique env vars you could use, there is no single yes/no this is Netlify CI, which this adds)

bcomnes (approved):

bcomnes:
I'm going to modify this implementation.

bkuhlmann:
🙇 @bcomnes. By the way, what is the ETA for when `3.3.0` becomes the standard build image for all builds as I currently see only `3.2.2` being used for current builds.

bcomnes:
I'm going to try to launch it today.

bkuhlmann:
🙇 @bcomnes for releasing `3.3.0`. I'm currently testing the `3.3.1` builds and not seeing the `NETLIFY=true` environment pair show up in the builds. In fact, I don't see the `NETLIFY` key in the output at all. When running the builds, I've been using the following Ruby code to get a pretty print of the environment:

``` ruby
pp ENV
```

I see all of the standard [Netlify Environments Variables](https://www.netlify.com/docs/continuous-deployment/#environment-variables) but don't see the new `NETLIFY` key.

Am I missing something?

bcomnes:
Apologies, I see the mistake.  I will try to correct tomorrow.

PR #300, opened 2019-04-20
Add rsvg.

kevincox:
The rsvg-convert command is a common way to rasterize images.


bcomnes (approved):
Thanks, seems reasonable.

PR #301, opened 2019-04-23
Wasmer support.

calavera:
This installs Wasmer in the base image, and uses Wapm to install
WebAssembly packages if a repository includes a wapm file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -422,14 +422,18 @@ ENV GIMME_GO_VERSION "1.12"
 ENV GIMME_ENV_PREFIX "/opt/buildhome/.gimme/env"
 RUN gimme
 
-
+################################################################################
+#
+# Wasmer
+#
+################################################################################
+RUN curl https://get.wasmer.io -sSfL | sh
 
 ################################################################################
 #
 # Dotnet Core
 #
 ################################################################################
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -422,14 +422,18 @@ ENV GIMME_GO_VERSION "1.12"
...
+#
+################################################################################
+RUN curl https://get.wasmer.io -sSfL | sh
 
 ################################################################################
 #
 # Dotnet Core
 #
 ################################################################################
-USER buildbot
```
bcomnes, line 432: Is this a mistake?

bcomnes (approved):

bcomnes:
I'll ping you when I stage it for testing.

Dockerfile
calavera: no, it's not necessary. The current user at that point is already buildbot, so this instruction was adding an extra layer without any reason.

bcomnes:
Going to get some pending changes out first, stage and test this.   Once its confirmed working, we can merge into xenial.

bcomnes:
Looks good.  It may require some more tweaking but we can land this for now.

bcomnes:
Thank you @calavera !

PR #307, opened 2019-04-30
Set wapm in the path, even when there is no wapm config files

calavera:
So you can install individual modules without having to configure them


bcomnes:
Ah sweet, you beat me to it.

bcomnes (approved):

PR #308, opened 2019-05-01
Handle ZOLA_VERSION

Keats:
Zola is already added in binrc but it looks like I missed the fact I needed to update this as well


bcomnes (approved):
LGTM

PR #309, opened 2019-05-01
Add Prep functions build function

bcomnes:
Adds a prep functions step to run_build functions, which we can run to prepare functions with zip-it-and-ship-it from inside the build environment, enabling fancier build and dependency installs. 

This is an alternative implementation from having separate buildbot stages, so that we can improve our bundling experience with the users build environment.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -483,7 +483,7 @@ install_dependencies() {
       exit 1
     fi
   fi
-  
+
   # zip-it-and-ship-it
   if [ -n "$ZISI_VERSION" ]
   then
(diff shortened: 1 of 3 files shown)


run-build-functions.sh
```diff
@@ -683,8 +711,3 @@ report_lingering_procs() {
     echo -e "${NC}"
   fi
 }
-
-after_build_steps() {
```
bcomnes, line 708: This just confuses things

bcomnes:
This is ready to go.

run-build-functions.sh
```diff
@@ -704,8 +732,3 @@ report_lingering_procs() {
...
-  # Find lingering processes after the build finished and report it to the user
-  report_lingering_procs
```
sgirones, line 710: we are no longer reporting on lingering_procs?

run-build-functions.sh
```diff
@@ -684,6 +684,34 @@ set_go_import_path() {
...
+  fi
+    echo Prepping functions with zip-it-and-ship-it $(zip-it-and-ship-it --version)
+
+    if zip-it-and-ship-it $functionsDir $zisiTempDir
+    then
+      echo "Prepping functions complete"
+      else
+        echo "Error prepping functions"
+        exit 1
+      fi
```
sgirones: indentation seems a bit hard to read. The most common way to write the conditional is:
```
if ...; then
  foo
else
  bar
fi
```

example:
```
if zip-it-and-ship-it $functionsDir $zisiTempDir; then
    echo "Prepping functions complete"
else
    echo "Error prepping functions"
    exit 1
fi
```

run-build-functions.sh
bcomnes: We do, but I just moved it into the run-build scripts.  It's not needed for local builds.

run-build-functions.sh
bcomnes: Thank you, I'm not great at bash.

sgirones: no one is good at bash ™️

bcomnes: OK this should be good now.

PR #315, opened 2019-05-14
Update docker pull command instructions

alexford:
Update the `docker pull` command instructions to suggest `build:latest`, because that's what the `start-image` script will attempt to grab when run from `master`. This avoids downloading `xenial` only to have the `start-image` script re-download `latest`.

Alternatively, it seems like perhaps all of Step 1 could be removed as the `start-image` script will fetch the image if it isn't available.


kitop:

Issue #318, opened 2019-05-29
Caching no longer works


yacoob:
I'm probably missing something here :)

I took a stab at adding Rust support to the build image. Things seem to be working fine, except my cache (as in, `$NETLIFY_CACHE_DIR`) is not getting populated. It looks like since 23b55c2959ea7a2b87742e7da7812514fb386bd0 `cache_artifacts` is not getting called at all during `run-build.sh`. As a result, no caching is performed and every build starts from scratch. Adding `cache_artifacts` as a last step before exiting from `run-build.sh` results in cache getting populated.

Now, the confusing bit is the fact that my actual deploys on netlify say:

```
10:44:16 PM: build-image tag: v3.3.2
```

but then continue to say:

```
10:44:47 PM: Caching artifacts
```

3.3.2 should have caching broken. I'm *obviously* missing something here, right? :D 

Thanks!

yacoob:
For reference: this is what I arrived at:
https://github.com/netlify/build-image/compare/xenial...yacoob:rust?expand=1

The caching wouldn't work in local tests, until I've added that explicit call to `cache_artifacts`.

XhmikosR:
This is definitely a regression in the Xenial image. Can we get a fix please?

PR #320, opened 2019-06-04
Add Rust support.

yacoob:
Stable toolchain only for the time being. Caching of `target/` dir is conditional on this directory actually being ignored in the repository. I've tested both cases with a [dummy rust project](https://github.com/yacoob/aoc-2018) and it seems to be working.

Note: this PR also adds `cache_artifacts` to `run-build.sh`, as it seems to have been removed recently, and without it caching doesn't happen at all (bug [here](https://github.com/netlify/build-image/issues/318)). Given the caching still happens in prod, maybe this simply got moved somewhere outside of this repository - if that's the case, feel free to remove this file from the PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -453,6 +453,14 @@ ENV DOTNET_ROOT "/opt/buildhome/.dotnet"
 RUN dotnet new
 WORKDIR /
 
+################################################################################
+#
+# Rust toolchain
+#
+################################################################################
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH "$PATH:/opt/buildhome/.cargo/bin"
+
 # Cleanup
 USER root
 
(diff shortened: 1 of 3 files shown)


yacoob:
Hello, anyone here? :)

leo60228:
Shouldn't you add at least one of the web targets? wasm32-unknown-unknown is recommended for new projects, but asmjs-unknown-emscripten and wasm32-unknown-emscripten are sometimes used for older ones. I believe the emscripten runs require emscripten to be installed manually, however.

yacoob:
I honestly have no idea about WASM. I only needed it for using a custom rust-based static site generator. I'd prefer to land this first, and leave WASM support for someone who knows what they're doing :D

leo60228:
Just do `rustup target add wasm32-unknown-unknown`.

arya-k:
I would also like to see this merged if possible - my build pipeline is in Rust

yacoob:
Is there anything going to happen here, or should I just revert this? :)

yacoob:
Allright, thank you for your attention. :[

jamen:
Could this be reopened? Netlify would be great for self hosting Rust docs, e.g. a `cargo doc` build command and using the `./target/doc` directory.

JeanBarriere:
Would also love a Rust support to support my wasm build in my webpack config :pray:

PR #324, opened 2019-07-05
Support specifying docker image version

mckelvaney:
Support the `NETLIFY_IMAGE` environment variable when running `./test-tools/start-image.sh` as per `./test-tools/start-image.sh`. This allows better control of local debugging of builds.


Baasie:
I seem to have a problem with the new version 3.3.5 in my site deploy.
I will try to check locally with docker if the version seems to be the problem, however, this will help me tremendously to see if the new version is the problem of the deploy.

I run webpack with fable and it seems to not run my webpack config anymore on 3.3.5.

kitop:

PR #326, opened 2019-07-29
Update run-build-functions.sh

Auspicus:
Gatsby stores it's build cache in the `.cache` folder. This PR might need some extra work to make sure the build is running with a Gatsby environment but if we know that there's no reason not to store the .cache directory. Gatsby handles invalidation. This would save Netlify from doing CPU intensive tasks repeatedly for every build like cropping images via gatsby-image. Might be able to borrow some logic from: https://github.com/netlify/build-image/compare/xenial...yacoob:rust?expand=1


gpoole:
A general-purpose cache folder to put cacheable build assets in would be great, doesn't even have to be just for Gatsby. Is this likely to be merged or are there issues with this approach?

kitop:

PR #329, opened 2019-08-12
Run buildbot v2 netlify-build step during build stage if feature flagged

shayonj:
#### Summary

This is basically ensuring the same/similar settings are being applied
in build-image's run-build script, compared to the one in buildbot's run-build
script

Build image: https://github.com/netlify/build-image/pull/329/files
Buildbot: https://github.com/netlify/buildbot/pull/445
Bitballoon: https://github.com/netlify/bitballoon/pull/4375

Brief plan/rollout steps: https://www.notion.so/netlify/Tech-Plan-Rolling-out-netlify-build-under-feature-flag-087bbb676c1c4c67a8d1bf2ed1c5808c#0c593fe4dc864cbfb2b5a86788c0c8d5

Github issue: https://github.com/netlify/product/issues/223

#### Tried locally

image


test-tools/start-image.sh
```diff
@@ -11,7 +11,7 @@ docker run --rm -t -i \
 	-e HUGO_VERSION \
 	-e PHP_VERSION \
 	-e GO_VERSION \
-	-v ${REPO_PATH}:/opt/repo \
+	-v ${REPO_PATH}:/opt/buildhome/repo \
```
shayonj, line 14: This was to ensure it works locally too.

keiko713:
@shayonj a bit confused reviewing this since the base branch is staging. can we maybe change PR to do the base branch as master?

shayonj:
So, build image doesn't have a master branch. Its either `xenial` or `trusty`, which are promoted from the image that was build using the `staging` branch. Ideally, we'd roll it out where we'd build the staging image, try it out against a feature flagged site and any other test scenario and then accordingly promote the images.

However, since build-image is sort of an "immutable infrastructure" relatively harder to iterate on, for the interim we are not baking in the package into this, but will during run time for feature flagged sites. We will bake it in once happy. 

Hence, closing it for now :D. Updated repos in others place to reflect this. Thanks for the 👀

keiko713:
ah, totally forgot about it 🤦‍♂ thanks for the context!

Issue #331, opened 2019-08-29
Bundler Version warning with no fix in sight


vphilot:
Can't finish build due to Bundle version -- follower instructions with no success, log below:

7:12:09 PM: Build ready to start
7:12:12 PM: build-image version: 8e315e54bc4032a32e73290be556cde4f8348c12
7:12:12 PM: build-image tag: v2.8.2
7:12:12 PM: buildbot version: b925d11411cabfe7b120501027bd0e96dbc28dde
7:12:13 PM: Fetching cached dependencies
7:12:13 PM: Starting to download cache of 302.7MB
7:12:18 PM: Finished downloading cache in 5.188965822s
7:12:18 PM: Starting to extract cache
7:12:22 PM: Finished extracting cache in 4.650002988s
7:12:23 PM: Finished fetching cache in 10.17806011s
7:12:23 PM: Starting to prepare the repo for build
7:12:23 PM: Preparing Git Reference refs/heads/master
7:12:33 PM: Starting build script
7:12:33 PM: Installing dependencies
7:12:35 PM: Downloading and installing node v10.16.3...
7:12:35 PM: Downloading https://nodejs.org/dist/v10.16.3/node-v10.16.3-linux-x64.tar.xz...
7:12:35 PM: 
#################
7:12:35 PM:                                    24.0%
7:12:35 PM: 
##################################################################
7:12:35 PM:  91.9%
7:12:35 PM: #
7:12:35 PM: #########################################
7:12:35 PM: ############################## 100.0%
7:12:35 PM: Computing checksum with sha256sum
7:12:35 PM: Checksums matched!
7:12:39 PM: Now using node v10.16.3 (npm v6.9.0)
7:12:39 PM: Attempting ruby version 2.6.2, read from environment
7:12:41 PM: Using ruby version 2.6.2
7:12:41 PM: Using PHP version 5.6
7:12:41 PM: Started restoring cached ruby gems
7:12:41 PM: Finished restoring cached ruby gems
7:12:41 PM: Installing gem bundle
7:12:42 PM: Warning: the running version of Bundler (2.0.1) is older than the version that created the lockfile (2.0.2). We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
7:12:45 PM: Fetching gem metadata from https://rubygems.org/
7:12:45 PM: .
7:12:46 PM: .
7:12:46 PM: .
7:12:46 PM: .
7:12:46 PM: .
7:12:46 PM: .
7:12:47 PM: .
[...]

Issue #328, opened 2019-08-05
Could not find 'bundler' (2.0.2)


igoynawamreh:
Jekyll deployment with bundler `v2.0.1` works fine.

I'm trying to update bundler to `v2.0.2`, I get this error:

```
9:55:36 AM: Build ready to start
9:55:38 AM: build-image version: 9e0f207a27642d0115b1ca97cd5e8cebbe492f63
9:55:38 AM: build-image tag: v3.3.2
9:55:38 AM: buildbot version: ef8d0929ed0baabafd8bbb7d0b021e1fc24180c0
9:55:39 AM: Fetching cached dependencies
9:55:39 AM: Starting to download cache of 119.4MB
9:55:39 AM: Finished downloading cache in 908.054542ms
9:55:39 AM: Starting to extract cache
9:55:43 AM: Finished extracting cache in 3.867926401s
9:55:43 AM: Finished fetching cache in 4.835648508s
9:55:43 AM: Starting to prepare the repo for build
9:55:44 AM: Preparing Git Reference refs/heads/master
9:55:45 AM: Starting build script
9:55:45 AM: Installing dependencies
9:55:45 AM: Started restoring cached node version
9:55:49 AM: Finished restoring cached node version
9:55:50 AM: v10.16.1 is already installed.
9:55:51 AM: Now using node v10.16.1 (npm v6.9.0)
9:55:51 AM: Attempting ruby version 2.6.2, read from environment
9:55:52 AM: Using ruby version 2.6.2
9:55:52 AM: Using PHP version 5.6
9:55:52 AM: Started restoring cached ruby gems
9:55:52 AM: Finished restoring cached ruby gems
9:55:52 AM: Installing gem bundle
9:55:53 AM: Warning: the running version of Bundler (2.0.1) is older than the version that created the lockfile (2.0.2). We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
9:55:55 AM: Fetching gem metadata from https://rubygems.org/
9:55:55 AM: .
9:55:55 AM: .
9:55:55 AM: .
9:55:55 AM: .
9:55:56 AM: .
9:55:56 AM: .
9:55:56 AM: .
9:55:56 AM: .
9:55:56 AM: .
9:55:56 AM: .
9:55:56 AM: .
9:55:57 AM: Fetching public_suffix 3.1.1
9:55:57 AM: Installing public_suffix 3.1.1
9:55:57 AM: Using addressable 2.6.0
[...]

Issue #296, opened 2019-04-08
respect "BUNDLED WITH" in Gemfile.lock to set bundler version


nolessafool:
Bundler versions are (sometimes?) recorded by bundler in the `Gemfile.lock`, for instance:

```
BUNDLED WITH
   1.12.5
```

Currently, we do not have a way to retroactively re-install and activate that version of bundler (vs our defaults) once we find it during dependency installation.  We'd need to detect and adjust before trying to process `Gemfile.lock` to support this workflow, which was requested by at least one paying customer who wishes to use a version other than the default.

Community:
- https://community.netlify.com/t/deployment-fails-for-latest-jekyll-4-0/4683/4
- https://community.netlify.com/t/building-a-jekyll-static-site-fails-due-to-ruby-env-versions/4963

<br><br><i>(related to <a href='https://netlify.zendesk.com/agent/tickets/10639'>Zendesk ticket #10639</a>)<br>gz#10639</i><br><br><i>(related to <a href='https://netlify.zendesk.com/agent/tickets/10850'>Zendesk ticket #10850</a>)<br>gz#10850</i><br><br><i>(related to <a href='https://netlify.zendesk.com/agent/tickets/10919'>Zendesk ticket #10919</a>)<br>gz#10919</i><br><br><i>(related to <a href='https://netlify.zendesk.com/agent/tickets/11307'>Zendesk ticket #11307</a>)<br>gz#11307</i>

nolessafool:
related issue:  https://github.com/netlify/build-image/issues/98

nolessafool:
customer to inform:  https://netlify.zendesk.com/agent/tickets/7282 and people in https://github.com/netlify/build-image/issues/328

Tickets:

https://netlify.zendesk.com/agent/tickets/7609
https://netlify.zendesk.com/agent/tickets/8109

Since this is the current version of bundler and we've gotten a half dozen reports already and the ONLY workaround is to downgrade to an older version of bundler locally, bumping priority to medium.

laurajodz:
Customer reported that removing "BUNDLED WITH" from Gemfile.lock got things working (https://netlify.zendesk.com/agent/tickets/7609)

soffes:
Related, if you specify a Ruby that isn't installed, it doesn't appear to install Bundler after installing your custom Ruby. It would be great if after checking your Ruby, it checked for `BUNDLED WITH` and then did `gem install -v bundler THAT_VERSION`.

shushugah:
[...]

PR #347, opened 2019-11-14
Pin bundler version so builds pinning it to old version keep working

Benaiah:
Pinning bundler to 2.0.1 was recommended as a workaround to not being able to install bundles created with the latest version of bundler. That was caused by the build-image only using a single
version of bundler and then going un-updated for a long time. This resulted in a lot of customers using Gemfile.locks with an old version of bundler (2.0.1) listed as BUNDLED_WITH. When the build-image
underlying the buildbot was updated and deployed, all these builds started failing because the build-image had the latest version of bundler (2.0.2).

See https://github.com/netlify/build-image/issues/328 and https://github.com/netlify/build-image/issues/331.

See also https://github.com/netlify/build-image/issues/296, which is the long-term solution to this issue.


Benaiah:
Fixing #296 seems to be pretty easy; closing this PR in favor of the proper fix.

PR #349, opened 2019-11-20
Fix test-tools/test-build.sh

Benaiah:
This script has either been broken for quite some time or never worked in the first place. Fixing it allows testing builds non-interactively. I based the changes to this script on the structure of `test-tools/start-image.sh`, which has been working the whole time.

If we decide not to merge this, we should remove `test-tools/test-build.sh` so it doesn't mislead people into thinking their build is broken when it's actually the test script.


Benaiah:
I'm going to go ahead and merge this at the end of my day; `test-tools/test-build.sh` is broken at the moment so this PR can't break anything further.

Issue #362, opened 2019-12-30
Support `PYTHON_VERSION`, for consistency


duckinator:
Of the [included languages](https://github.com/netlify/build-image/blob/xenial/included_software.md#languages) that have multiple versions available (Ruby, Node.js, Python, PHP, Go), only Python can not use a `<language name>_VERSION` environment variable.

It'd be super nice if this was supported, so it can be specified in `netlify.toml` like with the other languages.

duckinator:
I made an attempt at implementing this over at #363.

vbrown608:
Fixed by #363 - thanks again @duckinator !

PR #363, opened 2019-12-30
Add support for PYTHON_VERSION env variable.

duckinator:
My attempt at implementing #362.

I wasn't able to test it, unfortunately, because steps 2 and/or 3 in the README seem to be incorrect. There's no `build` command available in the Docker image.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -164,6 +164,17 @@ run_npm() {
   export PATH=$(npm bin):$PATH
 }
 
+check_python_version() {
+  if source $HOME/python${PYTHON_VERSION}/bin/activate
+  then
+    echo "Python version set to ${PYTHON_VERSION}"
+  else
+    echo "Error setting python version from $1"
+    echo "Please see https://github.com/netlify/build-image/#included-software for current versions"
+    exit 1
+  fi
+}
+
 install_dependencies() {
   local defaultNodeVersion=$1
   local defaultRubyVersion=$2
@@ -177,20 +188,14 @@ install_dependencies() {
   if [ -f runtime.txt ]
   then
     PYTHON_VERSION=$(cat runtime.txt)
-    if source $HOME/python${PYTHON_VERSION}/bin/activate
-    then
-      echo "Python version set to ${PYTHON_VERSION}"
-    else
-      echo "Error setting python version from runtime.txt"
-      echo "Please see https://github.com/netlify/build-image/#included-software for current versions"
-      exit 1
-    fi
+    check_python_version "runtime.txt"
   elif [ -f Pipfile ]
   then
     echo "Found Pipfile restoring Pipenv virtualenv"
     restore_cwd_cache ".venv" "python virtualenv"
   else
-    source $HOME/python${defaultPythonVersion}/bin/activate
+    PYTHON_VERSION=$defaultPythonVersion
+    check_python_version "the PYTHON_VERSION environment variable"
   fi
 
   # Node version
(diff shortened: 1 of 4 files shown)


duckinator:
Resolved the conflicts caused by Swift support being added.

duckinator:
I also have no idea what's up with this:

Screenshot_20200429_210138

vbrown608:
@duckinator it means that Netlify's backend engineering team was automatically notified that we need to review your PR. I'll take a look and follow up soon. Thanks for your contribution!

run-build-functions.sh
```diff
@@ -184,7 +185,7 @@ install_dependencies() {
     echo "Found Pipfile restoring Pipenv virtualenv"
     restore_cwd_cache ".venv" "python virtualenv"
   else
-    source $HOME/python2.7/bin/activate
```
vbrown608, line 192: The Docker container this runs in only has python 2.7, 3.5, and 3.7 installed. If `PYTHON_VERSION` isn't one of those, we should do some error handling like we're doing  when the user tries to load the version from `runtime.txt`.

vbrown608 (changes requested):
This is a nice improvement, thanks! One small comment about error handling but otherwise looking good.

For local testing - I'm able to run the `build` command in the Docker image as described by the README, though it errors with the following because of https://github.com/netlify/build-image/issues/402 
```
Version '--no-progress' not found - try `nvm ls-remote` to browse available versions.
```
Opening a PR to address that, in the meantime you can edit the last line of `test-tools/start-image.sh` as the submitter suggests.

If you're still having trouble, could you try making sure you've pulled the most recent version of the `netlify/build:xenial` image? I'm also happy to test the PR locally before merging :)

duckinator:
Rebased off `xenial`; gonna try to address @vbrown608's review now.

duckinator:
@vbrown608 I think this is ready for another look! :slightly_smiling_face:

erans:
+1 on being able to specify PYTHON_VERSION as an env var and (even more important to me) - being able to support Python 3.8!

vbrown608 (approved):
Sorry for the delay, tested this in our build system and it's looking good. Thanks for your contribution!

erans:
> Sorry for the delay, tested this in our build system and it's looking good. Thanks for your contribution!

Does it mean you can just control it via the env var or does it also mean there is Python 3.8 support?

vbrown608:
This PR adds the ability to control the python version through an env var but it must still be a supported version. It doesn't add support for 3.8.

revolter:
This doesn't work :( I set `PYTHON_VERSION` to `3.7`, but it still uses `2.7`.

revolter:
Using

```toml
[build.environment]
  PYTHON_VERSION = "3.7"
```

in a `netlify.toml` file doesn't work either :(

https://app.netlify.com/sites/fastlane-docs-preview/deploys/61991e5e6689420008c5416c

duckinator:
@revolter the documentation says only 2.7 and 3.8 are supported: https://github.com/netlify/build-image/blob/focal/included_software.md#languages

It seems for Python this isn't installing the language runtime on-the-fly, it's letting you choose from two installed versions. And in this case, the docs imply 3.8 will be used by default. So just removing that entirely might be your best bet.

duckinator:
Although the fact that it claims to default to 3.8, but falls back to 2.7 if it can't find the specified version is... bizarre...

revolter:
> @revolter the documentation says only 2.7 and 3.8 are supported: focal/included_software.md#languages

This was merged in the branch `xenial`, which means that the correct documentation is this: https://github.com/netlify/build-image/blob/xenial/included_software.md#languages, so `3.7` is supported.

> Although the fact that it claims to default to 3.8, but falls back to 2.7 if it can't find the specified version is... bizarre...

Same thing as above. On Xenial, `2.7` is the default.

PR #370, opened 2020-02-16
change volumes to delegated

svdarren:
Docker on MacOS has a known performance limitation for mounted volumes.  This creates a performance issue for the build-image, as dependency installs will cause many mounted volume edits.  Docker addresses this with the delegated option on mounts.  It has not effect on other OSes.


kitop:

PR #375, opened 2020-02-27
Define NETLIFY_BUILD_NODE_VERSION and "nvm install" it

Benaiah:
Also adds `--no-progress` flag to `nvm install` commands to clean up logs.


Benaiah:
#376 should fix the CI failure on this PR.

ehmicky (approved):
Looks good. Just one detail: I don't think you need to do `nvm use 10`. When doing `nvm install --no-progress 10`, it automatically switches to Node 10 at the end. You might want to add a comment though to specify that the command is meant here to both install and switch version.

Benaiah:
Ok, removed the extraneous `nvm use 10`.

PR #376, opened 2020-02-27
Pin setuptools to <45 to allow installing pipenv on python 2.7

Benaiah:
Fixes CI failures currently affecting any build-image PRs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -304,6 +304,7 @@ RUN virtualenv -p python3.7 --no-site-packages /opt/buildhome/python3.7 && \
     /bin/bash -c 'source /opt/buildhome/python3.7/bin/activate' && \
     ln -nfs /opt/buildhome/python3.7 /opt/buildhome/python3.7.2
 
+RUN /opt/buildhome/python${PIPENV_RUNTIME}/bin/pip install "setuptools<45"
 RUN /opt/buildhome/python${PIPENV_RUNTIME}/bin/pip install pipenv
 
 USER root


vbrown608:
I'm not very familiar with the python ecosystem - could you explain what you understood the error to be and why this fixes it?

Benaiah:
https://github.com/pypa/virtualenv/issues/1493 is the underlying issue that this is taking care of. I'm not a Python expert either, but as I understand it `pipenv` depends on `virtualenv`, which in turn requires `setuptools`. `setuptools` recently released a new version `45.0.0` which drops support for Python 2. This causes the build to fail because `pipenv` ends up installing `setuptools>=45`, which raises an error if installed on Python 2. By explicitly installing a version of `setuptools` that supports Python 2, we avoid the build failure that occurs on the next line.

I'm not sure why `pip` is installing a version of `setuptools` that doesn't support the running version of Python.

Benaiah:
https://jenkins.ops.netlify.com/blue/organizations/jenkins/build-image/detail/PR-373/1/pipeline is an example of a CI build that failed because of this issue instead of an issue with the PR itself.

vbrown608 (approved):
Thanks for explaining, this seems reasonable. My read of the thread is that maybe some mirror for installing Python packages is not respecting the requirement that `virtualenv` should use `setuptools <45.0.0` with Python 2.7, in which case we won't need this forever.

vbrown608:
Thanks for tracking this down!

Issue #217, opened 2018-11-15
minor: tighten up logging


fool:
There are a few loglines in the output of every build we could safely skip IMNSHO.  Not saying don't run the commands, just outputting them doesn't serve a purpose I can understand (vs say showing what $PATH is or something ;))

```
8:29:08 PM: unset GOOS;
8:29:08 PM: unset GOARCH;
8:29:08 PM: export GOROOT='/opt/buildhome/.gimme/versions/go1.10.linux.amd64';
8:29:08 PM: export PATH="/opt/buildhome/.gimme/versions/go1.10.linux.amd64/bin:${PATH}";
8:29:08 PM: go version >&2;
8:29:08 PM: export GIMME_ENV='/opt/buildhome/.gimme/env/go1.10.linux.amd64.env';
```

aarushik93:
These log lines are coming from the script from travis CI we use to install go, [see here](https://github.com/travis-ci/gimme/blob/4d1d7250f02d103de60c7b91fe1df9dde4c64304/gimme#L372). 
Might not be worth trying to remove those logs, if it's low priority?

mheffner:
@aarushik93 So are we supposed to be piping those commands into a bash environment? I'm wondering if instead of removing them in that Travis CI file there's a way to not print them when piped?

nolessafool:
thinking out loud:  we could also fork that travis script, change it, and use our own copy, non?

mheffner:
@fool that's absolutely an option, but I think if there's any way we can avoid having to maintain a fork that's ideal. If everyone else (ie Travis) is simply echoing those commands to their output then maybe we need our own version, it seems odd that those shell commands are showing up in output though. I'd be curious if there was a way we were invoking it that were different than expected use.

aarushik93:
As far as I can see we're not actually piping these commands. 
Making commands not print out is just:
set +x, as far as I know. 
But some of the commands aren't working when I pipe them as well. eg. go version >&2
Since to pipe and execute it should be in quotes

aarushik93:
Also, maybe someone can provide me some background on the use of the script from travis as opposed to just installing go ourselves in the container? Maybe that's a possibility too - I believe that would also remove other unnecessary logs?

PR #389, opened 2020-04-09
Pipe echo commands to bash shell

aarushik93:
Related to issue [217](https://github.com/netlify/build-image/issues/217)
Clears up all the unneeded logs + runs the commands. 

Leaving in 'go version', because original script doesn't have it in qutoes, meaning the command as well as output shows up.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -444,7 +444,7 @@ ENV GOCACHE "/opt/buildhome/.gimme_cache/gocache"
 # Install the default version
 ENV GIMME_GO_VERSION "1.12"
 ENV GIMME_ENV_PREFIX "/opt/buildhome/.gimme/env"
-RUN gimme
+RUN gimme | bash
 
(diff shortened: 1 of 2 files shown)


mheffner (commented):
@aarushik93 I think there appear to be two locations we run this. The Dockerfile should only be invoked during our own Jenkins CI build of the "build image". My guess is that the output the user is seeing during their builds is coming from the second time we invoke it in the functions script: https://github.com/netlify/build-image/blob/0e6e2bdfd3acc89e803dd41ae521cb2f2ee21ba4/run-build-functions.sh#L586.

aarushik93:
ahh updated, thanks for pointing that out!

mheffner (approved):
Looks like this should work, it looks like often you can [eval](https://github.com/travis-ci/gimme#installation--usage) the output of this, but we source the env file right after instead.

PR #391, opened 2020-04-10
add asciidoctor to the build node, useful for hugo

siscia:
Addressing this one: https://community.netlify.com/t/provide-asciidoc-asciidoctor-on-the-builder-nodes/12133

PR #392, opened 2020-04-10
add asciidoctor

siscia:
same as #391 for trusty


mraerino:
we're not maintaining the trusty image any further.

Issue #393, opened 2020-04-14
Issues with `node-libpq`


ehmicky:
A user [reported](https://community.netlify.com/t/unable-to-publish-function-using-postgres/12293/2) the following build error:

```
gyp info it worked if it ends with ok
gyp info using node-gyp@5.1.0
gyp info using node@10.20.1 | linux | x64
gyp info find Python using Python version 2.7.12 found at "/opt/buildhome/python2.7/bin/python"
gyp http GET https://nodejs.org/download/release/v10.20.1/node-v10.20.1-headers.tar.gz
gyp http 200 https://nodejs.org/download/release/v10.20.1/node-v10.20.1-headers.tar.gz
gyp http GET https://nodejs.org/download/release/v10.20.1/SHASUMS256.txt
gyp http 200 https://nodejs.org/download/release/v10.20.1/SHASUMS256.txt
gyp info spawn /opt/buildhome/python2.7/bin/python
gyp info spawn args [ '/opt/buildhome/.nvm/versions/node/v10.20.1/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py',
gyp info spawn args 'binding.gyp',
gyp info spawn args '-f',
gyp info spawn args 'make',
gyp info spawn args '-I',
gyp info spawn args '/opt/build/repo/node_modules/libpq/build/config.gypi',
gyp info spawn args '-I',
gyp info spawn args '/opt/buildhome/.nvm/versions/node/v10.20.1/lib/node_modules/npm/node_modules/node-gyp/addon.gypi',
gyp info spawn args '-I',
gyp info spawn args '/opt/buildhome/.cache/node-gyp/10.20.1/include/node/common.gypi',
gyp info spawn args '-Dlibrary=shared_library',
gyp info spawn args '-Dvisibility=default',
gyp info spawn args '-Dnode_root_dir=/opt/buildhome/.cache/node-gyp/10.20.1',
gyp info spawn args '-Dnode_gyp_dir=/opt/buildhome/.nvm/versions/node/v10.20.1/lib/node_modules/npm/node_modules/node-gyp',
gyp info spawn args '-Dnode_lib_file=/opt/buildhome/.cache/node-gyp/10.20.1/<(target_arch)/node.lib',
gyp info spawn args '-Dmodule_root_dir=/opt/build/repo/node_modules/libpq',
gyp info spawn args '-Dnode_engine=v8',
gyp info spawn args '--depth=.',
gyp info spawn args '--no-parallel',
[...]

PR #394, opened 2020-04-14
Fix postgresql build

fstn:
Custom build image to be able to use pg and pg-native in build


keiko713:
Hi @fstn , thanks for submitting PR!
Could you please share some examples of how you would like to use pg and pg-native during the build?

keiko713:
ah nvm, @ehmicky told me that it's about https://github.com/netlify/build-image/issues/393

mheffner:
Hi @fstn, thank you for your submission. However, at the moment we've decided that we do not want to increase the image footprint with additional database drivers. We understand this may limit the use of certain build packages and we are continuing to look at alternative approaches to make this as flexible as possible for users.

PR #401, opened 2020-04-27
Fix "swiftenv install"

Benaiah:
Don't pass `PLATFORM`, `URL`, or `VERSION` environment variables to `swiftenv install`. `swiftenv` expects those environment variables to refer to `swiftenv` internals (see https://github.com/kylef/swiftenv/issues/170).

See also #364.


mraerino (approved):

mraerino:
i think we should get this out so master works properly again

PR #404, opened 2020-04-28
Add Lua support

Benaiah:
Installs `lua5.1` (which is also available as `lua`), `lua5.3`, and `luajit`.

Does not install [LuaRocks](https://luarocks.org/), Lua's community package manager. This is because it isn't clear how to install LuaRocks in a way that would work for more than one installed version of Lua. The current version of LuaRocks expects Lua 5.3, but the `lua` CLI that's installed by apt is Lua 5.1.

Also evaluated https://dhavalkapil.com/luaver/, but that doesn't appear to be currently maintained and is significantly more complicated than this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -134,6 +134,9 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         libxss1 \
         libxtst6 \
         libyaml-dev \
+        lua5.1 \
+        lua5.3 \
+        luajit \
         mercurial \
         nasm \
         openjdk-8-jdk \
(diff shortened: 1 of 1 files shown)


mraerino:
With #411 you could just use the homebrew formula: https://formulae.brew.sh/formula-linux/lua
`luarocks` is also available there.
We might also refrain from installing this in the image and just tell people to use a `Brewfile`?

vbrown608:
We're going to pause adding new dependencies to the build image while we review our handling of preinstalled dependencies.

ingride:
should this PR be closed if we don't plan to ship it?

PR #406, opened 2020-04-30
Update versions and remove unused NETLIFY_BUILD_NODE_VERSION

Benaiah:


Dockerfile
```diff
@@ -258,8 +258,8 @@ ENV PATH /usr/local/rvm/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/s
 
 # Match this set latest Stable releases we can support on https://www.ruby-lang.org/en/downloads/
 RUN /bin/bash -c "source ~/.rvm/scripts/rvm && \
-                  rvm install 2.6.2 && rvm use 2.6.2 && gem install bundler && \
-                  rvm use 2.6.2 --default && rvm cleanup all"
+                  rvm install 2.7.1 && rvm use 2.7.1 && gem install bundler && \
+                  rvm use 2.7.1 --default && rvm cleanup all"
```
mheffner: Could we use ENV's for these so it's clearer and de-duped?

run-build.sh
```diff
@@ -18,11 +18,11 @@ if [[ ! -d $NETLIFY_REPO_DIR ]]; then
...
 
-: ${NODE_VERSION="10"}
-: ${RUBY_VERSION="2.6.2"}
-: ${YARN_VERSION="1.13.0"}
+: ${NODE_VERSION="12.16.2"}
+: ${RUBY_VERSION="2.7.1"}
+: ${YARN_VERSION="1.22.4"}
 : ${PHP_VERSION="5.6"}
-: ${GO_VERSION="1.12"}
+: ${GO_VERSION="1.14.2"}
```
mheffner, line 25: nit: anyway to combine these version numbers with what the Dockerfile uses to reduce duplication? Minor concern though if complicated

mheffner:
Also, maybe I'm missing something...but this seems like a lot of changes at once. Could we split the upgrades so that we can roll them out incrementally and monitor for increased failure rates?

run-build.sh
Benaiah: These versions are here as examples for the user to change if they want a different version in their test build. I'm not sure how we could deduplicate these without making this process more confusing to end users. (Note that this script in this repo is only used for debugging by end users).

Benaiah:
@mheffner I can split these into separate PRs if you think that'll be safer. Releasing this shouldn't affect any existing sites, though, since they'll already have whatever their versions of these tools are pinned and cached. Until the corresponding changes in the other repo are released, new sites will also continue to use the previously installed versions - they'll be installed on first use by the functions in `run-build-functions.sh` and cached thereafter.

ehmicky:
Trying to track down `NETLIFY_BUILD_NODE_VERSION` because it is used in several other places:
  - in [`build-image` `Dockerfile`](https://github.com/netlify/build-image/blob/0749a9c850ad21598a9c66af181cce606a799e7c/Dockerfile#L288) (which this PR updates): to `nvm install` the Node.js version used by `@netlify/build` and `@netlify/config`
  - in [`buildbot` `Dockerfile`](https://github.com/netlify/buildbot/blob/a2ef3b861c1199807274f3a443bce85953d833b4/Dockerfile#L36): to do the same? Also used as the Node.js version of `npm` when installing `@netlify/build` and `@netlify/config` 
  - in [`buildbot` `stage_build_consume_config.go`](https://github.com/netlify/buildbot/blob/a2ef3b861c1199807274f3a443bce85953d833b4/bot/stage_build_consume_config.go#L29): validation check to confirm the variable exists
  - in [`build-image` `run-build-functions`](https://github.com/netlify/build-image/blob/0749a9c850ad21598a9c66af181cce606a799e7c/run-build-functions.sh#L201), to prevent removing the Node.js version used by `@netlify/build` and `@netlify/config`

Could you confirm if I understood the rationale correctly:
  - the `NETLIFY_BUILD_NODE_VERSION` removed by this PR is dead code because it is overridden by the buildbot's `Dockerfile`
  - the `build-image` should expose the default Node.js version as `NODE_VERSION` instead

Should we split the `NETLIFY_BUILD_NODE_VERSION` change to its own PR to properly make sure everything is good with the Node.js version of `@netlify/build`, `@netlify/config`, `build.command` and Build plugins?

Dockerfile
```diff
@@ -274,20 +275,19 @@ USER root
...
-ENV YARN_VERSION=1.13.0
-
-ENV NETLIFY_BUILD_NODE_VERSION="12.16.1"
+ENV YARN_VERSION=1.22.4
+ENV NODE_VERSION=12.16.2
 
 RUN /bin/bash -c ". ~/.nvm/nvm.sh && \
-         nvm install --no-progress $NETLIFY_BUILD_NODE_VERSION && \
-         nvm install --no-progress 10 && npm install -g sm grunt-cli bower elm@$ELM_VERSION && \
+         nvm install --no-progress $NODE_VERSION && npm install -g sm grunt-cli bower elm@$ELM_VERSION && \
```
ehmicky, line 290: Are we switching from Node.js `10` to Node.js `12` for the default Node version used by the build image and the `build.command`?

mheffner:
Yeah :+1: for rolling out the node update separately given it's impact on more of the build-plugins work. I think it would be ideal to roll out the other upgrades individually as well, but defer to others on risk factor.

Dockerfile
Benaiah: Yes, though this will not go into effect until the corresponding PR in bitballoon is released.

Dockerfile
```diff
@@ -274,20 +275,19 @@ USER root
...
-    git checkout v0.34.0 && \
+    git checkout v"$NVM_VERSION" && \
     cd /
 
 ENV ELM_VERSION=0.19.0-bugfix6
-ENV YARN_VERSION=1.13.0
-
-ENV NETLIFY_BUILD_NODE_VERSION="12.16.1"
+ENV YARN_VERSION=1.22.4
+ENV NODE_VERSION=12.16.2
```
soemo: Why do you not use the latest version 12.16.3?

soemo: Or even better find a way to use the latest version node version in the dockerfile.
My problem currently is, that we defined the node version with 12 in the .nvmrc file.
That means in the docker we install a node version and in the next step my Netlify build will install the latest node again.

soemo: Or another idea why do you not check if a NODE_VERSION env variable is set and use this value?

Benaiah:
I split these into separate PRs: #413, #420, #421, #422. A further PR to upgrade Node will follow after https://github.com/netlify/build-image/pull/418 is merged.

Issue #402, opened 2020-04-28
readme states to use build:xenial - script runs build:latest


nickteeple:
readme states to `docker pull build:xenial` and run start-image.sh. When I run start-image.sh, docker ignores build:xenial and pulls build:latest which has older software defaults like ruby 2.4 instead of 2.6

latest docker for windows
launching from within ubuntu 16.04 windows WSL

nickteeple:
this change forced xenial to be used

> $ tail test-tools/start-image.sh  -n1
>         netlify/build:xenial /bin/bash

vbrown608:
Nice catch, thanks for reporting!

PR #407, opened 2020-05-01
Default to xenial tag for testing

vbrown608:
As reported in #402, the README suggests pulling `netlify/buildbot:xenial`, but the test script runs a test build with `netlify/buildbot:latest`. The `xenial` tag is our most up-to-date image and should be used for testing by default.

I think we should also apply the `latest` tag to `xenial`, since that's our latest image.

### DIFF
--- a/test-tools/start-image.sh
+++ b/test-tools/start-image.sh
@@ -2,6 +2,7 @@
 
 BASE_PATH=$(pwd)
 REPO_PATH=$(cd $1 && pwd)
+: ${NETLIFY_IMAGE="netlify/build:xenial"}
 
 docker run --rm -t -i \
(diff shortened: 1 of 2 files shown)


mheffner (approved):
:paintbrush:

Benaiah:
Is this intended to replace https://github.com/netlify/build-image/pull/390? Should we close that in favor of this?

vbrown608:
@Benaiah that's a good question, I wondered a bit myself whether we should do one or the other. Seems @mheffner is in favor of being a bit more specific about which tag we care about, so I'm gonna go with both.

I'd also like to add instructions to the README about using a custom or local tag.

Issue #408, opened 2020-05-01
Update steps in the README to test any image tag locally


vbrown608:
Users should be able to build the image locally, tag that image, and then pass the tag to the `start-image` script.

PR #409, opened 2020-05-01
Tag xenial builds as latest

vbrown608:
For now, xenial represents our most up-to-date build image. We should follow Docker conventions and tag it as latest.

Application of the `latest` tag is based on https://www.jenkins.io/doc/book/pipeline/docker/. I'm not a jenkins expert so feedback is welcome!

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,6 +38,9 @@ pipeline {
           docker.withRegistry('https://index.docker.io/v1/', 'docker-hub-ci') {
             docker.image("netlify/build:${env.BRANCH_NAME}").push()
             docker.image("netlify/build:${env.GIT_COMMIT}").push()
+            if (env.BRANCH_NAME == 'xenial') {
+              docker.image("netlify/build:{$env.BRANCH_NAME}").push('latest')
+            }
           }
         }
       }


mheffner:
Could we stop referencing the `latest` tag in our CI instead and explicitly reference `xenial`? I always felt `latest` was somewhat a misnomer, so referencing directly by the branch/version we want might be clearer? :shrug:

Benaiah:
@mheffner in the buildbot we reference a specific tagged release, not `latest`. The problem with not keeping `latest` up to date is that it's the default tag used by the `docker` CLI if you don't specify a tag. This leads to people running `docker pull netlify/build` and getting an image that was significantly out of date. We've had multiple users run into this in the past.

@vbrown608 I tried doing this myself, but I was trying to get the latest tagged release to get pushed to `latest`, and couldn't find a good way to do that (there's `buildingTag()`, but how do you tell if it's the latest tag being built?). Basing it on the `xenial` branch means people will sometimes get changes that have been merged but not released, but since I couldn't find a good alternative this seems to be the best solution.

Benaiah (approved):

mheffner:
@Benaiah That's fair, I noticed the README mentions the explicit tags even. Can we update the `test-build.sh` script to use the xenial tag though? It looks like it's defaulting to `latest`.

vbrown608:
@mheffner that's handled by #408

PR #411, opened 2020-05-03
Support installing tools via homebrew

mraerino:
Adding support for Homebrew aka Linuxbrew: https://docs.brew.sh/Homebrew-on-Linux

This shall be amazing because:
- It supports installing A LOT of tools without sudo: https://formulae.brew.sh/formula-linux/
- People can use a `Brewfile` in their git to specify dependencies

I also hope this will allow us to remove some lesser-used tooling from the image and offer users to install it via Homebrew instead.
Good candidates for that:
- pandoc
- hugo
- doxygen
- jq
- optipng
- asciidoctor
- emacs
- gifsicle
- graphicsmagick
- graphviz
- imagemagick
- jpegoptim
- pngcrush
- rsync
- sqlite3
- swig

## Test plan

Has been verified locally with `Brewfile` and custom Brewfile name via `$HOMEBREW_BUNDLE_FILE`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,6 +48,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         emacs25-nox \
         esl-erlang \
         expect \
+        file \
         fontconfig \
         fontconfig-config \
         g++ \
@@ -468,7 +469,6 @@ ENV DOTNET_ROOT "/opt/buildhome/.dotnet"
 #populate local package cache
 RUN dotnet new
 
-
 ################################################################################
 #
 # Swift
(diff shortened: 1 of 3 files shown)


mheffner:
Nice, I hadn't heard of Homebrew for Linux before. :boom:

mraerino:
yup, i remembered that a german webhosting company offered it along with full-fledged ssh access on a shared host and how this is exactly what we need (hope you can google-translate it): https://blog.uberspace.de/brauen-oder-toasten/

mraerino:
this is now based on an older commit, because the swift commit fails all builds

mraerino:
This is an example build which uses this feature successfully: https://app.netlify.com/sites/edge-functions-ts/deploys/5eb20aaf6ddfc20006027e9d

mheffner:
What if we removed this from the included_software doc so that we made it clear it's not generally available, then we could continue testing without officially "supporting" it? Then we could decide on a release date if this is a pattern we like.

calavera:
> What if we removed this from the included_software doc so that we made it clear it's not generally available, then we could continue testing without officially "supporting" it? Then we could decide on a release date if this is a pattern we like.

I think that it's a good idea, or even mark it as experimental. This change as it is makes much more sense than other tools that we've merged before, and I don't think we should hold it back.

mraerino:
this is currently blocked on https://github.com/netlify/build-image/pull/401
once that is merged i'm happy to rebase and update notes

calavera (approved):

mraerino:
@netlify/backend are you happy to merge this change?

run-build-functions.sh
```diff
@@ -449,6 +450,13 @@ install_dependencies() {
     fi
   fi
 
+  # Homebrew from Brewfile
+  if [ -f Brewfile ] || [ ! -z "$HOMEBREW_BUNDLE_FILE" ]
```
mheffner, line 454: fwiw, `-n` should also be equivalent to `! -z`

Dockerfile
```diff
@@ -482,6 +482,20 @@ ENV PATH "$SWIFTENV_ROOT/bin:$SWIFTENV_ROOT/shims:$PATH"
 RUN swiftenv install ${NETLIFY_BUILD_SWIFT_VERSION}
 RUN swift --version
 
+################################################################################
+#
+# Homebrew
+#
+################################################################################
+USER buildbot
+RUN /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```
mheffner, line 491: Any concern that we aren't pulling from a static/stable tag?

mheffner (approved):
Yeah, let's get it out

Dockerfile
mraerino: that's a good point. let's do that as part of a general release

PR #412, opened 2020-05-07
Add caching Cypress binary folder

### LINKED ISSUE #380: Add caching Cypress binary folder
Hi,

If we try to build a repo that has Cypress NPM dependency, the `npm install` step takes several minutes because each install caches only NPM node modules, but Cypress also downloads its binary and usually caches it in `~/.cache/cypress`, similar to Yarn.

I think I could add caching this folder in https://github.com/netlify/build-image/blob/xenial/run-build-functions.sh which would enable every user to run Cypress tests on Netlify or at least avoid the penalty of downloading Cypress again and again (if package json hash changes).

## Workaround

For now one can place the Cypress binary into `./node_modules` folder where it is cached with other node modules by setting the environment variable on Netlify project deploy settings

```
CYPRESS_CACHE_FOLDER: ./node_modules/CypressBinary
```

guillett:
Fixes #380

Cypress redownload takes 1 minute on a 5-minute build.
cf. https://app.netlify.com/sites/mes-aides-front/deploys/5eb39dacc1004900062246ed

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -377,7 +377,7 @@ install_dependencies() {
   if [ -f requirements.txt ]
   then
     echo "Installing pip dependencies"
-    restore_home_cache ".cache" "pip cache"
+    restore_home_cache ".cache" "~/.cache"
     if pip install -r requirements.txt
     then
       echo "Pip dependencies installed"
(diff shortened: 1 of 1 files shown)


vbrown608:
Unfortunately because of the way the filesystem is organized in our build system, copying from the cache to `$home/.cache` doesn't have good performance. We made caching of that directory more restrictive with https://github.com/netlify/build-image/pull/443.

We're looking at configuring Cypress to store the binary in the `node_modules` directory, which we cache by default.

PR #417, opened 2020-05-11
Do not run `yarn` if `package.json` has not changed

ehmicky:
This is part of https://github.com/netlify/buildbot/issues/735

We currently skip `npm install` when the `package.json` has not changed between builds. We should do the same with `yarn`.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -118,6 +118,8 @@ run_yarn() {
       echo "Error during Yarn install"
       exit 1
     fi
+
+    echo "$(shasum package.json)-$NODE_VERSION" > $NETLIFY_CACHE_DIR/package-sha
   fi
   export PATH=$(yarn bin):$PATH
 }
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -98,23 +98,26 @@ run_yarn() {
   fi
 
 
-  echo "Installing NPM modules using Yarn version $(yarn --version)"
-  run_npm_set_temp
-
-  # Remove the cache-folder flag if the user set any.
-  # We want to control where to put the cache
-  # to be able to store it internally after the build.
-  local yarn_local="${YARN_FLAGS/--cache-folder * /}"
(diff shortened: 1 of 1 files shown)


run-build-functions.sh
```diff
@@ -424,7 +427,7 @@ install_dependencies() {
     swiftenv rehash
     echo "Finished restoring cached Swift version"
   fi
-  
+
```
ehmicky, line 432: This whitespace keeps being prettified in my IDE, I think it's probably using a tab instead of spaces.

run-build-functions.sh
```diff
@@ -98,23 +98,26 @@ run_yarn() {
   fi
 
 
-  echo "Installing NPM modules using Yarn version $(yarn --version)"
-  run_npm_set_temp
+  if install_deps package.json $NODE_VERSION $NETLIFY_CACHE_DIR/package-sha
```
ehmicky, line 101: This is the same line of code that we currently use for `npm install`.

run-build-functions.sh
vbrown608: Looks to me like there was some trailing whitespace on this line that your editor cleaned up. Good with me!

run-build-functions.sh
vbrown608: It's interesting that we're running this against `package.json` instead of a lockfile if one is present.

I think mirroring the behavior of `run_npm` is the right idea for now, but let's keep an eye on it. Could imagine an edge case where someone changed `yarn.lock` but not `package.json`.

vbrown608 (approved):
Nice catch! :yarn:

run-build-functions.sh
ehmicky: Yes, I noticed this too.
Same here: I was thinking mirroring would be best for the moment, but in the future we might consider using `package-lock.json` or `yarn.lock` (if they exist, since those are optional).

PR #413, opened 2020-05-07
Remove NETLIFY_BUILD_NODE_VERSION & special handling of its nvm dir

Benaiah:
This is now installed in a separate location in a different layer of the build image. The `NETLIFY_BUILD_NODE_VERSION` behavior contained here is nonfunctional.

I moved this change out of https://github.com/netlify/build-image/pull/406 so it could be evaluated independently.

PR #418, opened 2020-05-13
Move NF-build node to later image and lock node 10 version.

mheffner:
Preparing for future netlify-build image changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -273,18 +273,23 @@ USER root
 
 RUN curl -o- -L https://yarnpkg.com/install.sh > /usr/local/bin/yarn-installer.sh
 
+ENV NVM_VERSION=0.35.3
+
 # Install node.js
 USER buildbot
 RUN git clone https://github.com/creationix/nvm.git ~/.nvm && \
     cd ~/.nvm && \
-    git checkout v0.34.0 && \
+    git checkout v$NVM_VERSION && \
     cd /
 
 ENV ELM_VERSION=0.19.0-bugfix6
 ENV YARN_VERSION=1.13.0
 
+ENV NETLIFY_NODE_VERSION="10.20.1"
+
 RUN /bin/bash -c ". ~/.nvm/nvm.sh && \
-         nvm install --no-progress 10 && npm install -g sm grunt-cli bower elm@$ELM_VERSION && \
+         nvm install --no-progress $NETLIFY_NODE_VERSION && \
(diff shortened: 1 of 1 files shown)


vbrown608:
> Move NF-build node to later image

I don't see where we're installing the node version for Netlify Build now. According to https://github.com/netlify/build/issues/1239#issuecomment-626032865 it looks like we want 12.16.*. Am I missing it?

ehmicky:
I think it is installed in the [buildbot `Dockerfile`](https://github.com/netlify/buildbot/blob/0d19834c19ebfe54208907d729718495339fe085/Dockerfile#L23).

vbrown608:
Oh, nice!

vbrown608 (approved):
:art:

mheffner:
Yeah, we're going to pull the NF node into a separate build image layer that we'll install the plugins in. Since we'll need to update that more rapidly, figured separating from the build-image would be better.

Benaiah (approved):

Dockerfile
```diff
@@ -273,21 +273,23 @@ USER root
...
     cd ~/.nvm && \
-    git checkout v0.34.0 && \
+    git checkout v$NVM_VERSION && \
     cd /
 
 ENV ELM_VERSION=0.19.0-bugfix6
 ENV YARN_VERSION=1.13.0
 
-ENV NETLIFY_BUILD_NODE_VERSION="12.16.1"
+ENV NETLIFY_NODE_VERSION="10.20.1"
```
Benaiah, line 288: Why is this `NETLIFY_NODE_VERSION` and not `NODE_VERSION`?

mheffner: No particular reason, more that it represents the version Netlify has chosen for the image. :shrug:

ehmicky: Note that, as opposed to `NETLIFY_NODE_VERSION`, `NODE_VERSION` is the name of the environment variable used by users to specify a custom Node.js version. I am not sure whether that would be a good thing or not in this instance.

mheffner: Yeah, that shouldn't interfere. These ENV's are only present during the docker build phase.

Benaiah: Ok, that makes sense.

ehmicky (approved):

mheffner:
This branch name is somewhat a misnomer now that https://github.com/netlify/build-image/pull/413 has been merged. This just locks the Node version and upgrades NVM version slightly.

PR #421, opened 2020-05-14
Upgrade preinstalled Go version to 1.14.4

Benaiah:
Split off from https://github.com/netlify/build-image/pull/406.


erquhart:
Need a review here when someone has time 🙏 cc/ @mheffner @vbrown608

mheffner (approved):
:shipit:

mheffner:
actually, looks like there may be a few more minor versions if we want to grab the very latest

Benaiah:
@mheffner good catch; updated the PR to use `1.14.4`.

mheffner (approved):

PR #424, opened 2020-05-15
Cache `.netlify/plugins/` directory

ehmicky:
Part of https://github.com/netlify/build/issues/1281

When Build plugins cannot be found in the `build-image` pre-installed cached, and were not added to the `package.json`, `@netlify/build` automatically installs those.

At the moment, we install those in the build's main `node_modules`, but this creates many caching issues. Instead, we will now install those in `.netlify/plugins/node_modules`. A dummy `.netlify/plugins/package.json` is also created. 

Therefore `.netlify/plugins` should now be cached in the `build-image`. This PR implements this.


vbrown608 (commented):
:pineapple:

vbrown608 (approved):

PR #430, opened 2020-05-21
Fix logs when caching `~/.cache`

### LINKED ISSUE #353: ~/.cache is not only the pip cache
The `~/.cache` directory is treated as the pip cache, but it's actually more than that. It's the default value for `$XDG_CACHE_HOME` according to the XDG Base Directory specification, so there are probably many more tools that use this directory. In fact, yarn places its cache in this directory by default, for example.

The current build script always stores it in the cache, but it only restores it if there is a `requirements.txt` file found, so if the project is understood to be a Python project.

I think it's a great idea to cache this directory but it should also always be restored, and the logs should be changed to reflect that it's more than the pip cache.

https://github.com/netlify/build-image/blob/9cade8af58c2cf3a17a1e9433d2e979149488837/run-build-functions.sh#L621
https://github.com/netlify/build-image/blob/9cade8af58c2cf3a17a1e9433d2e979149488837/run-build-functions.sh#L364

ehmicky:
Fixes #353.

We are caching `~/.cache` but the logs show it as `pip cache`. `pip` is cached in `~/.cache/pip`.

I think it does make sense to cache anything in `~/.cache` as it is a convenient well-known cache location. Anything in there is intended for caching. It is [standardized as `$XDG_CACHE_HOME`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) by `freedesktop.org`. 

However we should rename it. This PR renames it to the path itself `~/.cache` since I am assuming this might be more precise and clearer to developers than `home cache`, `$HOME/cache`, `user cache` or `global cache`. But I am open to any suggestions, we just need another name than `pip`.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -374,7 +374,7 @@ install_dependencies() {
   if [ -f requirements.txt ]
   then
     echo "Installing pip dependencies"
-    restore_home_cache ".cache" "pip cache"
+    restore_home_cache ".cache" "~/.cache"
     if pip install -r requirements.txt
     then
       echo "Pip dependencies installed"
@@ -677,7 +677,7 @@ cache_artifacts() {
   cache_cwd_directory ".netlify/plugins" "build plugins"
 
   cache_home_directory ".yarn_cache" "yarn cache"
-  cache_home_directory ".cache" "pip cache"
+  cache_home_directory ".cache" "~/.cache"
   cache_home_directory ".cask" "emacs cask dependencies"
   cache_home_directory ".emacs.d" "emacs cache"
   cache_home_directory ".m2" "maven dependencies"


guillett:
:+1: for the provided details but please consider #412 instead as it also allow caching that folder in a node build :) :pray:  Many thanks!

ehmicky:
Oh, sorry about this, I did not notice #412. Those are indeed almost identical.

What about splitting this into 2 PRs? This PR could only update the log name.

Your PR could be more about where `~/.cache` is being cached/restored. At the moment, it is only restored when there is a `requirements.txt`. Your PR also restores it when there is a `package.json`. But since `~/.cache` is generic, I am guessing it should ways be cached/restored?

vbrown608:
Closing this PR as we're now only caching ~/.cache/pip and not the full ~/.cache directory, so the log line is correct.

PR #439, opened 2020-06-17
Preinstall Ruby 2.6.2

vbrown608:
Many users are still pinned to Ruby 2.6.2, and are seeing an error like
ruby
Although we're defaulting to 2.7 for new sites, we should continue to preinstall 2.6.2 or most users builds will be slowed down by about 15 seconds, regardless of whether they're using Ruby.


mheffner (commented):
Do we need to do the gem bundle install for that version as well?

vbrown608:
That's a good point. Buildbot will install Bundler if it's missing (and it installs a lot faster than Ruby) but I think it makes sense to go ahead and preinstall it if we're preinstalling the Ruby version :+1:

mheffner (approved):
:hamburger:

Issue #196, opened 2018-09-04
Support for monorepos, including lerna or yarn workspaces


herkulano:
Currently, you're only caching the `node_modules` at the root level. 
With lerna or yarn workspaces, there are sub-folders with `node_modules` in them that are not cached. This causes a failure to build. It then builds correctly if it's retried with a clean cache.

fool:
related internal issue:  https://github.com/netlify/buildbot/issues/163

zanona:
For my use-case (yarn workspaces) I have been able to work around this issue by using:

**netlify.toml**
```toml
[build]
  command = "npm i yarn@1.10.1 && yarn && yarn build"
```
Of course, this is not ideal as it takes additional time to run npm install and yarn install again, however the cache dir which the npm installed version of yarn uses is the same so it should be all ready for the next time a build is triggered. Bottom line, it has increase building times by 1 minute for me.

Hopefully we will be able to revert to the expected behaviour soon, though :)

Btw, thanks for the help @fool

eckdanny-osi:
Strange. I have a `YARN_VERSION` env var in my web config (set to `1.9.4`) and [the build logs](https://app.netlify.com/sites/cws-cares/deploys/5bc1499e82d3f15e9e16f9b6) seem to indicate its working...

but i changed my `package.json#scripts#build` to `"build": "yarn -v && exit 1"` and the result is *not* expected: 

```log
...
6:32:06 PM: Executing user command: yarn build
6:32:06 PM: yarn run v1.9.4
6:32:06 PM: $ yarn -v && exit 1
6:32:06 PM: 1.3.2
```

The issue I'm seeing in my dead builds:

`error An unexpected error occurred: "patterns.map is not a function".`

was resolved by https://github.com/yarnpkg/yarn/pull/5114 and should work since [yarn v1.4](https://github.com/yarnpkg/yarn/releases/tag/v1.4.0)

I'll try @zanona's workaround.

*Why do i not have the yarn version i think i should have?*

iddan:
I expect Netlify to be able to look for different Netlify.toml files: So for deploying a monorepo with multiple sites I would create multiple Netlify sites each of them with a different Netlify.toml.

xiaoxiangmoe:
add YARN_VERSION to env and it will support yarn workspaces

```toml
[build.environment]
  YARN_VERSION = "1.13.0"
```

tsaiDavid:
[...]

PR #440, opened 2020-06-18
Allow to force Yarn usage

Embraser01:
This PR add a new environment variable that allows anyone to force the usage of yarn. It's a simple workaround of #196.

The current workaround for Yarn workspaces ([here](https://github.com/netlify/build-image/issues/196#issuecomment-562328172)) only works with Yarn v1, Yarn v2 use yarn.lock to determine the root of the project and so, setting a empty `yarn.lock` in a specific folder will fail.

I choose the env var `USE_YARN` but it could be anything really 😄.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -458,7 +458,7 @@ install_dependencies() {
   if [ -f package.json ]
   then
     restore_cwd_cache node_modules "node modules"
-    if ([ -f yarn.lock ] && [ ! "$USE_YARN" = "false" ]) || [ "$USE_YARN" = "true" ]
+    if [ "$USE_YARN" = "true" ] || ([ "$USE_YARN" != "false" ] && [ -f yarn.lock ]) 
     then
       run_yarn $YARN_VERSION
     else
@@ -471,7 +471,7 @@ install_dependencies() {
   then
     if ! [ $(which bower) ]
     then
-      if ([ -f yarn.lock ] && [ ! "$USE_YARN" = "false" ]) || [ "$USE_YARN" = "true" ]
+      if [ "$USE_YARN" = "true" ] || ([ "$USE_YARN" != "false" ] && [ -f yarn.lock ]) 
       then
         echo "Installing bower with Yarn"
         yarn add bower
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -458,7 +458,7 @@ install_dependencies() {
   if [ -f package.json ]
   then
     restore_cwd_cache node_modules "node modules"
-    if [ -f yarn.lock ] || [ ! -z $USE_YARN ]
+    if ([ -f yarn.lock ] && [ ! "$USE_YARN" = "false" ]) || [ "$USE_YARN" = "true" ]
     then
       run_yarn $YARN_VERSION
     else
@@ -471,7 +471,7 @@ install_dependencies() {
   then
     if ! [ $(which bower) ]
     then
-      if [ -f yarn.lock ] || [ ! -z $USE_YARN ]
+      if ([ -f yarn.lock ] && [ ! "$USE_YARN" = "false" ]) || [ "$USE_YARN" = "true" ]
       then
         echo "Installing bower with Yarn"
         yarn add bower
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -458,7 +458,7 @@ install_dependencies() {
   if [ -f package.json ]
   then
     restore_cwd_cache node_modules "node modules"
-    if [ "$USE_YARN" = "true" ] || ([ "$USE_YARN" != "false" ] && [ -f yarn.lock ]) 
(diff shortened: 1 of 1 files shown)


ehmicky:
Hi @Embraser01,

Thanks a lot for working on this!
This makes sense to me. 
@mheffner @vbrown608 What do you think?

Note: we would also need to:
  - Update `@netlify/build` since it runs `yarn` detecting `yarn.lock` as well
  - Add to the documentation website

Embraser01:
> Note: we would also need to:
> 
> * Update `@netlify/build` since it runs `yarn` detecting `yarn.lock` as well

Looking at the code for `@netlify/build`, it seems to be a little trickier to make it work with Yarn v2:

Some options have changed, `--no-progress` and `--no-interactive` don't exist anymore so [this command](https://github.com/netlify/build/blob/8ab25f493e1ac4d7401af71b917440e296a1d88a/packages/build/src/install/main.js#L58) will fail...

Furthermore, the `--cache-folder` option is not used anymore (there is actually a special case for Netlify to report it as a warning instead of an error [here](https://github.com/yarnpkg/berry/blob/b79663659e6f9275e66d7c458b34f0e5c140e675/packages/plugin-essentials/sources/commands/install.ts#L79-L108)). The global cache folder used by Yarn 2 is `~/.yarn/berry/cache/` but can be moved with the `YARN_GLOBAL_FOLDER` env var ([ref.](https://yarnpkg.com/configuration/yarnrc#globalFolder)).

run-build-functions.sh
```diff
@@ -458,7 +458,7 @@ install_dependencies() {
   if [ -f package.json ]
   then
     restore_cwd_cache node_modules "node modules"
-    if [ -f yarn.lock ]
+    if [ -f yarn.lock ] || [ ! -z $USE_YARN ]
```
ehmicky: Could you check that the value of this variable is `true`?
Like this, this allows for more values (including `false`) to be used as well.

ehmicky:
Thanks for all that information! I have summarized it in a separate issue for `@netlify/build` only at https://github.com/netlify/build/issues/1535. Could you please check it out and confirm I haven't missed anything?

Any help to improve Yarn support in the `build-image` is really appreciated!

run-build-functions.sh
Embraser01: Ok, I changed it to check only for `false` or `true`

run-build-functions.sh
```diff
@@ -458,7 +458,7 @@ install_dependencies() {
   if [ -f package.json ]
   then
     restore_cwd_cache node_modules "node modules"
-    if [ -f yarn.lock ]
+    if ([ -f yarn.lock ] && [ ! "$USE_YARN" = "false" ]) || [ "$USE_YARN" = "true" ]
```
```suggestion
    if [ "$USE_YARN" = "true" ] || ([ "$USE_YARN" != "false" ] && [ -f yarn.lock ]) 
```

Small performance optimization since checking for file existence is slower than doing a string comparison. So if `USE_YARN` is either `true` or `false`, `&& [ -f yarn.lock ]` will be short-circuited.

run-build-functions.sh
```diff
@@ -471,7 +471,7 @@ install_dependencies() {
   then
     if ! [ $(which bower) ]
     then
-      if [ -f yarn.lock ]
+      if ([ -f yarn.lock ] && [ ! "$USE_YARN" = "false" ]) || [ "$USE_YARN" = "true" ]
```
```suggestion
      if [ "$USE_YARN" = "true" ] || ([ "$USE_YARN" != "false" ] && [ -f yarn.lock ]) 
```

ehmicky (approved):

ehmicky:
This looks good to me!
@vbrown608 @mheffner What do you think?

vbrown608:
Looks good. I suggest we prefix the environment variable like `NETLIFY_USE_YARN` to avoid the kind of collisions we saw with `SITE_ID`.

(It would be weird if someone has `USE_YARN` set and doesn't want to use yarn, but we might as well make it a practice going forward.)

run-build-functions.sh
```diff
@@ -458,7 +458,7 @@ install_dependencies() {
   if [ -f package.json ]
   then
     restore_cwd_cache node_modules "node modules"
-    if [ -f yarn.lock ]
+    if [ "$USE_YARN" = "true" ] || ([ "$USE_YARN" != "false" ] && [ -f yarn.lock ])
```
```suggestion
    if [ "$NETLIFY_USE_YARN" = "true" ] || ([ "$NETLIFY_USE_YARN" != "false" ] && [ -f yarn.lock ]) 
```

run-build-functions.sh
```diff
@@ -471,7 +471,7 @@ install_dependencies() {
   then
     if ! [ $(which bower) ]
     then
-      if [ -f yarn.lock ]
+      if [ "$USE_YARN" = "true" ] || ([ "$USE_YARN" != "false" ] && [ -f yarn.lock ])
```
```suggestion
      if [ "$NETLIFY_USE_YARN" = "true" ] || ([ "$NETLIFY_USE_YARN" != "false" ] && [ -f yarn.lock ]) 
```

ehmicky:
Issue to add to documentation: https://github.com/netlify/docs/issues/632

PR #443, opened 2020-06-30
Don't cache full .cache dir

vbrown608:
Pip caches files to `$home/.cache/pip`. In order to take advantage of the Pip cache, we cache all of `$home/.cache` at the end of each build.

At the start of the next build, we conditionally restore the `.cache` directory to the home directory if `requirements.txt` is present.

This has two problems:

1. We are caching everything in `.cache`, but only restoring it under certain conditions. This can mean uploading many files to the cache without any benefit.
2. Because of the way the filesystem for builds is organized, restoring large files from the Netlify cache to the home directory is slow. We want to avoid restoring many/large files to the home directory if possible.

This PR caches only the `pip` subdirectory of `.cache`.

~Marking WIP for testing.~

Internal issue: https://github.com/netlify/buildbot/issues/845

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -46,7 +46,7 @@ mkdir -p $NETLIFY_CACHE_DIR/.netlify/plugins
 
 # HOME caches
 mkdir -p $NETLIFY_CACHE_DIR/.yarn_cache
-mkdir -p $NETLIFY_CACHE_DIR/.cache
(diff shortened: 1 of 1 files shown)


ehmicky:
Related PR (which might need to be closed): https://github.com/netlify/build-image/pull/430

ehmicky (approved):

PR #449, opened 2020-07-14
feat: support pnpm

### LINKED ISSUE #447: Feature request: Add PNPM support
**Which problem is this feature request solving?**

[pnpm](https://pnpm.js.org/) is a robust, mature dependency manager for Node.js projects, and is much faster than npm and yarn. It elegantly [solves npm doppelgängers problem and has stricter module resolution, solving yarn ghost modules](https://rushjs.io/pages/maintainer/package_managers/).

Currently, Netlify users need to use [workarounds](https://community.netlify.com/t/using-pnpm-and-pnpm-workspaces/2759/4) in order to use pnpm, involving altering a proper project configuration and making it less optimal.

**Describe the solution you'd like**

Build images should have `pnpm@latest` preinstalled as a global npm module.
If a `pnpm-lock.yaml` file is detected in the `build.base` folder, `pnpm install --frozen-lockfile` should be run instead of `npm install` (or even better, the _install command_ should just be exposed as a configuration value so that people can run what they want instead of `npm install`).

**Can you submit a pull request?**

~Yes~ ~No~ Maybe.

jcayzac:
Fixes #447

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -46,6 +46,7 @@ mkdir -p $NETLIFY_CACHE_DIR/.netlify/plugins
 
 # HOME caches
 mkdir -p $NETLIFY_CACHE_DIR/.yarn_cache
+mkdir -p $NETLIFY_CACHE_DIR/.pnpm-store
 if [[ "$NETLIFY_CACHE_PIP_SUBDIRECTORY" == "true" ]]; then
   mkdir -p $NETLIFY_CACHE_DIR/.cache/pip
 else
@@ -61,6 +62,7 @@ mkdir -p $NETLIFY_CACHE_DIR/.gimme_cache/gocache
 mkdir -p $NETLIFY_CACHE_DIR/.wasmer/cache
 
 : ${YARN_FLAGS=""}
+: ${PNPM_FLAGS=""}
 : ${NPM_FLAGS=""}
 : ${BUNDLER_FLAGS=""}
 
@@ -124,6 +126,47 @@ run_yarn() {
   export PATH=$(yarn bin):$PATH
 }
 
+run_pnpm() {
+  local pnpm_version=$1
+  restore_home_cache ".pnpm-store" "pnpm store"
+
+  if [ $(which pnpm) ] && [ "$(pnpm --version)" != "$pnpm_version" ]
+  then
+    echo "Found pnpm version ($(pnpm --version)) that doesn't match expected ($pnpm_version)"
+    npm uninstall pnpm -g
+  fi
+
+  if ! [ $(which pnpm) ]
+  then
+    echo "Installing pnpm at version $pnpm_version"
+    npm install -g pnpm@$pnpm_version
+  fi
+
+  echo "Installing NPM modules using PNPM version $(pnpm --version)"
+  run_npm_set_temp
+
+  # Remove the store-dir flag if the user set any.
+  # We want to control where to put the cache
+  # to be able to store it internally after the build.
+  if [ -f .npmrc ]
+  then
+    sed -i '/store-dir/d' .npmrc
+  fi
+  local pnpm_local="${PNPM_FLAGS/--store-dir * /}"
+  # The previous pattern doesn't match the end of the string.
+  # This removes the flag from the end of the string.
+  pnpm_local="${pnpm_local%--store-dir *}"
+
+  if pnpm install --store-dir $NETLIFY_BUILD_BASE/.pnpm-store ${pnpm_local:+"$pnpm_local"}
+  then
+    echo "NPM modules installed using PNPM"
+  else
+    echo "Error during PNPM install"
+    exit 1
+  fi
+  export PATH=$(pnpm bin):$PATH
+}
+
 run_npm_set_temp() {
   # Make sure we're not limited by space in the /tmp mount
   mkdir $HOME/tmp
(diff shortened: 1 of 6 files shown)


run-build-functions.sh
```diff
@@ -79,7 +81,7 @@ install_deps() {
 }
 
 run_yarn() {
-  yarn_version=$1
+  local yarn_version=$1
```
ehmicky: This changes seems unrelated to pnpm, would it be possible to remove it?

ehmicky:
Thanks a lot for this PR @jcayzac!

run-build-functions.sh
jcayzac: Sure

jcayzac:
@ehmicky what repo got the source of https://docs.netlify.com/configure-builds/manage-dependencies/ ? The docs also need a PR

ehmicky:
Good catch @jcayzac! The docs are a private repository.
Putting @verythorough and @klavavej in the loop since they are working on the documentation website.

run-build-functions.sh
```diff
@@ -124,6 +126,47 @@ run_yarn() {
...
+
+  echo "Installing NPM modules using PNPM version $(pnpm --version)"
+  run_npm_set_temp
+
+  # Remove the store-dir flag if the user set any.
+  # We want to control where to put the cache
+  # to be able to store it internally after the build.
+  if [ -f .npmrc ]
+  then
+    sed -i '/store_dir/d' .npmrc
```
jcayzac: Oops, typo, it should be `store-dir` and not `store_dir`.

jcayzac: Fixed

ehmicky:
For the record, it appears that very few Netlify users are using pnpm: only 13 accounts in the last 7 days.

jcayzac:
> For the record, it appears that very few Netlify users are using pnpm: only 13 accounts in the last 7 days.

Maybe because Netlify doesn't support it? Out of curiosity, how did you collect that number?
My PNPM deploys for instance have no package.json and use Makefiles, to avoid any automatic install with npm/yarn.

antfu:
Come from [this thread](https://community.netlify.com/t/using-pnpm-and-pnpm-workspaces/2759/6) and I am using pnpm as well. Would love to see pnpm get supported by Netlify!

BTW, https://github.com/netlify/cli/pull/550 seems to only merge with the Girdsome changes but not for `pnpm` as its title says. I guess we can bring back these changes https://github.com/netlify/cli/pull/550/commits/131bd5ed4e7121ffdd42f5e2aa6948ea0837c4b3 once we make this PR merged and working.

erquhart:
@jcayzac I have a bit of bad news - we're going to ask that you take a different route for using pnpm on Netlify. I do realize this PR was invited, but on further consideration, there's some concern around the added complexity of natively supporting an additional package manager, a support effort that doesn't end with the initial implementation.

What we would _rather_ see is a route by which users of various tools like pnpm can self implement. We've played with supporting homebrew as one option for this (which we plan to revisit), and build plugins may be another. 

Current recommendation would be to utilize the approach outlined in the [community post](https://community.netlify.com/t/using-pnpm-and-pnpm-workspaces/2759/3), which is known to be effective for production use cases.

All of that said, we _do_ want to keep the discussion open toward a solution, but I expect that solution will likely be through a build plugin. Build plugins don't quite have all of the capabilities necessary to fully support pnpm, but @ehmicky may have some ideas.

Conversation to continue in https://github.com/netlify/build/issues/1633.

jcayzac:
@erquhart 

> What we would rather see is a route by which users of various tools like pnpm can self implement. 

This would actually be easy if the run build script didn't "get in the way", i.e. if we could disable it entirely so that e.g. `npm` doesn't run when we have a `package.json`.

mcmxcdev:
There are like 3 main players in the dependency management ecosystem now, so I cannot understand the reasoning for "added complexity" above.

Both npm and yarn are not viable alternatives for us, so this is a big disappointment.

Hacky workarounds might work for a while, but then again, already encountering issues to due non-straightforward support of pnpm: https://github.com/netlify/netlify-plugin-nextjs/issues/219

Maybe this decision can be reconsidered as pnpm steadily increases its user base.

ZetiMente:
> @jcayzac I have a bit of bad news - we're going to ask that you take a different route for using pnpm on Netlify. I do realize this PR was invited, but on further consideration, there's some concern around the added complexity of natively supporting an additional package manager, a support effort that doesn't end with the initial implementation.
> 
> What we would _rather_ see is a route by which users of various tools like pnpm can self implement. We've played with supporting homebrew as one option for this (which we plan to revisit), and build plugins may be another.
> 
> Current recommendation would be to utilize the approach outlined in the [community post](https://community.netlify.com/t/using-pnpm-and-pnpm-workspaces/2759/3), which is known to be effective for production use cases.
> 
> All of that said, we _do_ want to keep the discussion open toward a solution, but I expect that solution will likely be through a build plugin. Build plugins don't quite have all of the capabilities necessary to fully support pnpm, but @ehmicky may have some ideas.
> 
> Conversation to continue in [netlify/build#1633](https://github.com/netlify/build/issues/1633).

pnpm has emerged as even more popular than yarn. Has Netlify changed their mind yet? Like I can use npm but its just way slower than pnpm. And pnpm is practically required for large monorepos. Tis a shame.

jwarby:
Just to add to the discussion - we're trying to have a go at switching from Yarn Classic + Lerna to pnpm and we're getting stuck with Netlify builds too. The workaround in the community post doesn't work for us as we have a Netlify build plugin that comes from a private package, and the Netlify build will fail if this plugin is not installed before the build process kicks off proper. This means we will have to let Netlify install dependencies first, then we'll have to cleanup and re-install using pnpm which is a big waste of resources and adds further complexity to our processes which we'd rather avoid.

PR #450, opened 2020-07-16
Move build binaries from /usr/local/bin to /opt/build/bin

Benaiah:
This will remove them from the `PATH`, and hopefully reduce the number of builds that accidentally call the build binaries. A corresponding PR in buildbot is at https://github.com/netlify/buildbot/pull/861.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -485,8 +485,9 @@ WORKDIR /
 USER root
 
 # Add buildscript for local testing
-ADD run-build-functions.sh /usr/local/bin/run-build-functions.sh
-ADD run-build.sh /usr/local/bin/build
+RUN mkdir -p /opt/build/bin
+ADD run-build-functions.sh /opt/build/bin/run-build-functions.sh
+ADD run-build.sh /opt/build/bin/build
 ADD buildbot-git-config /root/.gitconfig
 RUN rm -r /tmp/*
 
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -788,7 +788,7 @@ set_go_import_path() {
 }
 
 find_running_procs() {
-  ps aux | grep -v [p]s | grep -v [g]rep | grep -v [b]ash | grep -v "/usr/local/bin/buildbot" | grep -v "/opt/build/bin/buildbot" | grep -v [d]efunct | grep -vw '\[build\]'
+  ps aux | grep -v [p]s | grep -v [g]rep | grep -v [b]ash | grep -v "/opt/build/bin/buildbot" | grep -v [d]efunct | grep -vw '\[build\]'
 }
 
 report_lingering_procs() {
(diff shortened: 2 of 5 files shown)


run-build-functions.sh
```diff
@@ -788,7 +788,7 @@ set_go_import_path() {
 }
 
 find_running_procs() {
-  ps aux | grep -v [p]s | grep -v [g]rep | grep -v [b]ash | grep -v "/usr/local/bin/buildbot" | grep -v [d]efunct | grep -vw '\[build\]'
+  ps aux | grep -v [p]s | grep -v [g]rep | grep -v [b]ash | grep -v "/usr/local/bin/buildbot" | grep -v "/opt/build/bin/buildbot" | grep -v [d]efunct | grep -vw '\[build\]'
```
ehmicky: Should `grep -v "/usr/local/bin/buildbot"` be removed?

run-build-functions.sh
```diff
@@ -466,7 +466,7 @@ install_dependencies() {
   if [ -f package.json ]
```
ehmicky, line 466: Should `/usr/local/bin/yarn-installer.sh` be moved as well?
It is also in `Dockerfile`.

ehmicky (approved):

Benaiah:
@ehmicky I've updated the PR to respond to your suggestions. Thanks for the review!

Dockerfile
```diff
@@ -293,10 +291,13 @@ ENV YARN_VERSION=1.22.4
...
 
-RUN /bin/bash -c ". ~/.nvm/nvm.sh && \
+RUN /bin/bash -c "mkdir /tmp/yarn-installer && \
+         curl -o- -L https://yarnpkg.com/install.sh > \"/tmp/yarn-installer/yarn-installer.sh\" && \
+         . ~/.nvm/nvm.sh && \
          nvm install --no-progress $NETLIFY_NODE_VERSION && \
          npm install -g sm grunt-cli bower elm@$ELM_VERSION && \
-             bash /usr/local/bin/yarn-installer.sh --version $YARN_VERSION && \
+         bash /tmp/yarn-installer/yarn-installer.sh --version $YARN_VERSION && \
+         rm -rf /tmp/yarn-installer && \
```
ehmicky: `yarn-installer.sh` seems to be used here as well: 

https://github.com/netlify/build-image/blob/ca811f47d4c1cbd1812d1eb6ecb0c977e86d1a1d/run-build-functions.sh#L100

Removing it would prevent executing it later?

Benaiah: Oh, good catch. Maybe we shouldn't move this to its own directory, then, since it will be needed at build time. Either way, I think we can deal with that separately from this PR. I'm going to remove the `yarn-installer.sh` logic from this PR.

PR #460, opened 2020-07-27
Move build binaries again (to /opt/build-bin this time)

Benaiah:
`/opt/build/bin` was being overwritten in production, so it wasn't working as a location for the build binaries. It had to be reverted due to that. This PR moves them instead to `/opt/build-bin`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -485,10 +485,9 @@ WORKDIR /
 USER root
 
 # Add buildscript for local testing
-ENV BUILD_BIN_DIR "/opt/build-bin"
-RUN mkdir -p ${BUILD_BIN_DIR}
-ADD run-build-functions.sh ${BUILD_BIN_DIR}/run-build-functions.sh
-ADD run-build.sh ${BUILD_BIN_DIR}/build
+RUN mkdir -p /opt/build-bin
+ADD run-build-functions.sh /opt/build-bin/run-build-functions.sh
+ADD run-build.sh /opt/build-bin/build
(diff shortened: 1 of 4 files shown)


vbrown608 (approved):
Added a subjective comment, up to you if you wanna make the change :+1:

Benaiah:
@vbrown608 I'm not seeing your comment anywhere, did it get deleted?

vbrown608:
*shakes fist at github*

Dockerfile
```diff
@@ -485,8 +485,10 @@ WORKDIR /
 USER root
 
 # Add buildscript for local testing
-ADD run-build-functions.sh /usr/local/bin/run-build-functions.sh
-ADD run-build.sh /usr/local/bin/build
+ENV BUILD_BIN_DIR "/opt/build-bin"
+RUN mkdir -p ${BUILD_BIN_DIR}
+ADD run-build-functions.sh ${BUILD_BIN_DIR}/run-build-functions.sh
+ADD run-build.sh ${BUILD_BIN_DIR}/build
```
vbrown608: In my opinion, we're making too much use of environment variables to dry up scripts and Dockerfiles. While we get rid of some repetition, we lose the ability to glance at a line and see exactly what's happening.

Given that we don't currently write unit tests for things like `run-build-functions.sh`, I think we should keep them more explicit and un-code-like.

Benaiah: Sounds good to me - I was iffy on this anyway. I'll change it back to use the paths themselves.

PR #461, opened 2020-07-27
Fix find_running_procs to exclude correct binary

Benaiah:
This should have been in https://github.com/netlify/build-image/pull/460 but I missed it. This script is overridden in production, so this isn't very important, but it is a little misleading as is.


vbrown608 (approved):
Nice catch! Sorry I missed it in review.

Benaiah:
np! thanks for the quick review on this

PR #463, opened 2020-07-30
Cache cypress binary under node_modules

vbrown608:
Proposal to cache the Cypress binary under `node_modules`, which we persist to the build cache by default.

This configuration is [recommended](https://github.com/cypress-io/netlify-plugin-cypress#recommended) in the docs for netlify-plugin-cypress. This would turn it on for all users.

This is preferable to caching the default location of the Cypress binary (~/.cache) because it avoids copying across the partitions that exist in the build system.


vbrown608:
Actually, we need to set this later, closing this PR.

PR #465, opened 2020-08-03
Fix support for Yarn 2 (berry)

### LINKED ISSUE #319: Yarn arguments
Opening an issue here as requested in the support email (I initially asked on Twitter what was the best PoC, and was redirected to the support) 🙂

Yarn 2 doesn't support the `--cache-folder` command-line option (it always was problematic, because users had to specify it at every single command). This causes a crash on Netlify when using Yarn together with the v2 (currently flagged experimental, fortunately).

Additionally, starting from Yarn 2 we no longer have different logic between the Yarn cache and the [offline mirrors](https://yarnpkg.com/blog/2016/11/24/offline-mirror/). The directory pointed by the `cache-folder` key will be populated with archives, and as such it becomes possible (and recommended in various situations) to add the cache to the repository (amongst other things, it enables [Zero-installs](https://yarnpkg.github.io/berry/features/zero-installs)). However, this doesn't work if Netlify configures a different cache location than the default one.

We're currently hitting this issue with the Yarn 2 Netlify instance ([example](https://app.netlify.com/sites/keen-leavitt-98f35d/deploys/5cf41659a550ce000bd435ab)), but it's likely your customers would encounter it as well once we start publishing formal releases. I'd be happy to help find out a solution (I lead Yarn's development) 🙂

ylemkimon:
- Checks the version of Yarn installed globally using [`YARN_IGNORE_PATH=1`](https://yarnpkg.com/configuration/yarnrc#ignorePath). The environment variable `YARN_VERSION` sets the version of Yarn to install globally.
- Sets [`YARN_GLOBAL_FOLDER`](https://yarnpkg.com/configuration/yarnrc#globalFolder) environment variable to the cache directory. The `cache-folder` flag is deprecated in Yarn 2. The user can opt-in to use the global cache by setting [`YARN_ENABLE_GLOBAL_CACHE`](https://yarnpkg.com/configuration/yarnrc#enableGlobalCache) in Netlify settings. (See https://github.com/netlify/build-image/pull/465#discussion_r465539682)

Closes #319.

**Suggestion for Docs:**
https://docs.netlify.com/configure-builds/environment-variables/#netlify-configuration-variables
```diff
- * `YARN_VERSION`: Sets the [Yarn version].
+ * `YARN_VERSION`: Sets the [Yarn version] to install globally.
+ * `YARN_ENABLE_GLOBAL_CACHE`: Use the [global cache](https://yarnpkg.com/configuration/yarnrc#enableGlobalCache) for Yarn 2. 
```

https://docs.netlify.com/configure-builds/manage-dependencies/#yarn
```diff
- * `YARN_VERSION`: defaults to the version preinstalled with your initial [build image]. Accepts any released version number.
+ * `YARN_VERSION`: defaults to the version preinstalled with your initial [build image]. Accepts any released version number. The specified version is installed globally.
+ * `YARN_ENABLE_GLOBAL_CACHE`: whether to use the [global cache](https://yarnpkg.com/configuration/yarnrc#enableGlobalCache) for Yarn 2. If you use zero-installs, do not set this variable.
```

CC @arcanis Could you check whether this is the right way to use Yarn 2 in CI?

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -113,8 +113,8 @@ run_yarn() {
   then
     # The cache-folder flag is deprecated in Yarn 2
     # Override the environment variable if the user set any.
+    # To use the global cache, the user should set YARN_ENABLE_GLOBAL_CACHE
     export YARN_GLOBAL_FOLDER="$NETLIFY_BUILD_BASE/.yarn_cache"
-    export YARN_ENABLE_GLOBAL_CACHE=1
   else
     # Remove the cache-folder flag if the user set any.
     yarn_local="${yarn_local/--cache-folder * /}"
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -85,7 +85,7 @@ run_yarn() {
   # Check the version of yarn installed globally
   if [ $(which yarn) ] && [ "$(YARN_IGNORE_PATH=1 yarn --version)" != "$yarn_version" ]
   then
-    echo "Found yarn version ($(yarn --version)) that doesn't match expected ($yarn_version)"
+    echo "Global yarn version ($(yarn --version)) that doesn't match expected ($yarn_version)"
     rm -rf $NETLIFY_CACHE_DIR/yarn $HOME/.yarn
     npm uninstall yarn -g
   fi
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -85,7 +85,7 @@ run_yarn() {
   # Check the version of yarn installed globally
   if [ $(which yarn) ] && [ "$(YARN_IGNORE_PATH=1 yarn --version)" != "$yarn_version" ]
   then
-    echo "Global yarn version ($(yarn --version)) that doesn't match expected ($yarn_version)"
+    echo "Global yarn version ($(yarn --version)) doesn't match expected ($yarn_version)"
     rm -rf $NETLIFY_CACHE_DIR/yarn $HOME/.yarn
     npm uninstall yarn -g
   fi
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -82,7 +82,8 @@ run_yarn() {
   fi
   restore_home_cache ".yarn_cache" "yarn cache"
 
-  if [ $(which yarn) ] && [ "$(yarn --version)" != "$yarn_version" ]
+  # Check the version of yarn installed globally
+  if [ $(which yarn) ] && [ "$(YARN_IGNORE_PATH=1 yarn --version)" != "$yarn_version" ]
   then
     echo "Found yarn version ($(yarn --version)) that doesn't match expected ($yarn_version)"
     rm -rf $NETLIFY_CACHE_DIR/yarn $HOME/.yarn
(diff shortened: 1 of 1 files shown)


run-build-functions.sh
```diff
@@ -98,19 +99,32 @@ run_yarn() {
...
-  local yarn_local="${YARN_FLAGS/--cache-folder * /}"
-  # The previous pattern doesn't match the end of the string.
-  # This removes the flag from the end of the string.
-  yarn_local="${yarn_local%--cache-folder *}"
+  if [[ "${yarn_version%%.*}" -gt "1" ]]
+  then
+    # The cache-folder flag is deprecated in Yarn 2
+    # Override the environment variable if the user set any.
+    export YARN_GLOBAL_FOLDER="$NETLIFY_BUILD_BASE/.yarn_cache"
+    export YARN_ENABLE_GLOBAL_CACHE=1
```
arcanis: I'm not a fan of using environment variables for this, because it'll break people using zero-installs (when the cache is inside the repository), overriding the config they explicitly requested (I know it's what the code already did, overriding `--cache-folder` etc, but that really strikes me as a bad idea).

I think a more gentle approach would be to create a `.yarnrc.yml` file in the home directory that would contain the "default options" you want to have. This way, users would still be able to opt-out of the default workflow to use their own.

run-build-functions.sh
```diff
@@ -82,7 +82,8 @@ run_yarn() {
   fi
   restore_home_cache ".yarn_cache" "yarn cache"
 
-  if [ $(which yarn) ] && [ "$(yarn --version)" != "$yarn_version" ]
+  # Check the version of yarn installed globally
+  if [ $(which yarn) ] && [ "$(YARN_IGNORE_PATH=1 yarn --version)" != "$yarn_version" ]
   then
     echo "Found yarn version ($(yarn --version)) that doesn't match expected ($yarn_version)"
```
arcanis:
```suggestion
    echo "Global yarn version ($(yarn --version)) that doesn't match expected ($yarn_version)"
```

arcanis (commented):
Thanks for doing this! I've left a few comments, I hope someone from Netlify will be able to take a look.

run-build-functions.sh
ylemkimon: @arcanis Thank you for the review! I've removed `export YARN_ENABLE_GLOBAL_CACHE=1`. The user now can opt-in to use global cache by setting `YARN_ENABLE_GLOBAL_CACHE` in Netlify settings.

I guess it'd be for Netlify to decide the default for `enableGlobalCache` and whether to allow users to override the global folder, so I'll leave it for now.

kara-todd: Chiming in here... :)

I've also attempted to use zero-installs on Netlify and ran into issues as the install command [always seems to run](https://community.netlify.com/t/skip-yarn-install-when-using-yarn-2-with-zero-install/19423/4). Would it be possible to amend these changes to allow skipping of the install step completely? Or did I just miss how to do that 🙃 ?

ylemkimon: @kara-todd AFAIK, shoudn't `yarn install` only take few seconds with zero-installs? You can set `YARN_ENABLE_NETWORK=0` to disallow making network requests.

kara-todd: @ylemkimon when I was running it... it still seemed to still run the relink step, but I didn't try it with `YARN_ENABLE_NETWORK=0`. It didn't take terribly long as I recall... 🤔  However, I use this in conjunction with Netlify CMS which has to wait on builds for content reviews... So every second ends up being precious. :)

I know that part of the problem was caused by Netlify not correctly detecting the cache directory as you are trying to fix here. Still, a true `zero install` where I only run my build command would be preferable to me if it's not a huge lift.

ylemkimon: @kara-todd I have a Netlify repo with zero-install and it takes less than two seconds running `yarn install`. If it takes more than a few seconds, the repo may not have been set up properly. A workaround for now would be move the whole project into a subdirectory, not having `package.json` and `yarn.lock` at the root, and `cd` in the build command.

benfurber:
Would love for this to be merged in so that I move a site over to netlify. 🙏

kevinji:
The line below `export PATH=$(yarn bin):$PATH` also needs to be updated; `yarn bin` on Yarn 2+ no longer returns a `node_modules/.bin` directory (which doesn't exist with PnP) but instead just [lists all available binaries](https://yarnpkg.com/cli/bin).

kitop:

Feel free to reopen against focal branch.

PR #470, opened 2020-09-02
Proposal: add some tests to our bash scripts

erezrokah:
Found this https://github.com/bats-core/bats-core while thinking about https://github.com/netlify/build-image/issues/196.

Not sure this is something we would like to add/invest time in or if there is a better approach.

In this PR you can see an example of testing the `run_yarn` command.

We would need to refactor some of our functions to be more "testable" (if possible).

Opening as draft to get some feedback.

### Update

Looks like the npm version is out of date with the one in GitHub:
https://github.com/bats-core/bats-core/releases
https://www.npmjs.com/package/bats

We can always install from source.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "main": "index.js",
   "scripts": {
     "version": "auto-changelog -p -t keepachangelog -l false --starting-commit 9d2e26a && git add CHANGELOG.md",
-    "release": "git push && git push --tags && gh-release"
+    "release": "git push && git push --tags && gh-release",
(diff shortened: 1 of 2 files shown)


ehmicky:
I think anything that's adding more tests to our buildbot is a huge win :+1: 

I would personally favor integration tests over module or unit tests, since they are closer to the end user experience, but testing our Bash logic is definitely better than not doing it :)

erezrokah:
Closing as stale

PR #471, opened 2020-09-03
Document `PYTHON_VERSION` option

verythorough:
Functionality added in https://github.com/netlify/build-image/pull/363

Contributes to the closure of https://github.com/netlify/docs/issues/616.

Adds `PYTHON_VERSION` as an option in the included software list, in the same format as other languages.


ingride (approved):
lgtm

PR #474, opened 2020-09-28
Support installing tools via Homebrew (take 2)

calavera:
@mraerino opened a PR to support homebrew long time ago (#411). We discovered that some people had Brewfile files in their sites and this integration broke their builds because we were trying to install packages that were not necessary.

Supporting homebrew is a great idea, and I'd like to bring this feature back. In order to not break builds again, I renamed the `Brewfile` to `Brewfile.netlify`. I'd expect no breaking changes since I just made this up.


mraerino:
can we hide this whole process behind some environment variable?
that way it's probably safer for people working on the build process

run-build-functions.sh
```diff
@@ -449,6 +450,13 @@ install_dependencies() {
     fi
   fi
 
+  # Homebrew from Brewfile
+  if [ -f Brewfile.netlify ] || [ ! -z "$HOMEBREW_BUNDLE_FILE" ]
+  then
+    echo "Installing Homebrew dependencies from ${HOMEBREW_BUNDLE_FILE:-Brewfile.netlify}"
+    brew bundle --file Brewfile.netlify
```
mraerino: i'm not sure what happens here if you set `HOMEBREW_BUNDLE_FILE` to something custom.

what do you think about making `Brewfile.netlify` a default value for `HOMEBREW_BUNDLE_FILE` inside the if and remove the `--file`?

calavera: oh yeah.

depadiernos:
Is there anything needed here before we can review and merge?

included_software.md
```diff
@@ -72,3 +72,10 @@ The specific patch versions included will depend on when the image was last buil
 * [Doxygen](http://www.doxygen.org) - 1.8.6
 * [WASMER](https://github.com/wasmerio/wasmer)
 * [WAPM](https://github.com/wasmerio/wapm-cli)
```
mraerino: we'll want to remove these two lines if wasmer is really going

mraerino (commented):
i'm not too sure about removing wasmer in this PR.
I was not able to find any evidence of anyone using it in the last 7 days directly, but i'm unsure.

I'd defer this decision to the team

cecton:
This PR is blocking me personally for a month now (other people for longer). I think if people really want WASMER you either need to upgrade your ubuntu image or contact the people from WASMER and ask them to compile with an older version of GLIBC.

included_software.md
calavera: removed

calavera:
> i'm not too sure about removing wasmer in this PR.

We need to do it anyways, and this is the last PR that actually passes CI because wasmer doesn't work anymore, so I'm all for merging both and move on. It really have never been used.

mraerino (approved):
cool with me.

syrusakbary:
> We need to do it anyways, and this is the last PR that actually passes CI because wasmer doesn't work anymore, so I'm all for merging both and move on. It really have never been used.

Update: we have actually fixed Wasmer in `master`, it no longer have the glibc issue thanks to this PR:

https://github.com/wasmerio/wasmer/pull/1755

PS: if you need I can update the binary manually so it's also fixed when you install Wasmer `1.0.0-alpha4` (the latest published version).

I'd recommend you to keep Wasmer, super cool things are coming in the next weeks!

calavera:
> I'd recommend you to keep Wasmer, super cool things are coming in the next weeks!

people will be able to install wasmer when needed via homebrew, and update to newer versions when they want instead of depending on us to keep up with new releases.

syrusakbary:
> people will be able to install wasmer when needed via homebrew

Yeah, that's a great alternative.

Although being able to cache wasmer artifacts independently in-between builds will make a bit of a difference in the future!

cecton:
> people will be able to install wasmer when needed via homebrew, and update to newer versions when they want instead of depending on us to keep up with new releases.

I made #477 because I'm wasting a lot of build time downloading Rust at every build. Does installing via homebrew would solve that?

ingride:
@cecton thanks for adding the Rust PR, we can look into supporting that as part of our build image, but meanwhile, as soon as this PR is released, you'd be able to install Rust via homebrew which should help.

ingride (approved):
lgtm, thanks for making the change

PR #475, opened 2020-09-30
Remove lingering processes logic

ehmicky:
This removes the lingering processes logic from the `build-image`.

The `build-image` only defines the function. The `buildbot` used to do the actual call. Both have been moved to `@netlify/build` in https://github.com/netlify/buildbot/pull/979.

https://github.com/netlify/buildbot/pull/979 remove the `buildbot` call to those functions, so this PR only cleans them up (they are now dead code). This PR can be reviewed, but it would be good to wait few days after https://github.com/netlify/buildbot/pull/979 to be released in production first, so we are sure the new logic works and we don't have to rollback the `build-image` in case of problems.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -905,24 +905,3 @@ set_go_import_path() {
     cd $importPath
   fi
 }
-
-find_running_procs() {
-  ps aux | grep -v [p]s | grep -v [g]rep | grep -v [b]ash | grep -v "/opt/build-bin/buildbot" | grep -v [d]efunct | grep -vw '\[build\]'
-}
-
-report_lingering_procs() {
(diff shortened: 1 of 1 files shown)


vbrown608:
This is a thoughtful cleanup, thanks for including it.

Looks like wasmer is failing due to a missing dependency. We don't appear to be pinning the version. https://github.com/netlify/build-image/blob/ff8f5a3541b520f070dfd237a41cd9789de3b915/Dockerfile#L451

ehmicky:
I opened https://github.com/netlify/build-image/issues/476 for this problem.
This does not appear to be related to this PR though.

vbrown608 (approved):

Issue #476, opened 2020-10-01
CI builds are failing


ehmicky:
CI builds are failing due to some issue related to Wasmer.

[Example of failed build](https://jenkins.ops.netlify.com/blue/organizations/jenkins/build-image/detail/xenial/68/pipeline):

```
RUN curl https://get.wasmer.io -sSfL | sh
Welcome to the Wasmer bash installer!

downloading: wasmer-linux-amd64
Latest release: 1.0.0-alpha3
Downloading archive from https://github.com/wasmerio/wasmer/releases/download/1.0.0-alpha3/wasmer-linux-amd64.tar.gz
######################################################################## 100.0%

installing: /opt/buildhome/.wasmer
Updating bash profile /opt/buildhome/.profile
we've added the following to your /opt/buildhome/.profile
If you have a different profile please add the following:

# Wasmer
export WASMER_DIR="/opt/buildhome/.wasmer"
[ -s "$WASMER_DIR/wasmer.sh" ] && source "$WASMER_DIR/wasmer.sh"
/opt/buildhome/.wasmer/bin/wasmer: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.27' not found (required by /opt/buildhome/.wasmer/bin/wasmer)
error: wasmer was installed, but doesn't seem to be working :(
ERROR: executor failed running [/bin/sh -c curl https://get.wasmer.io -sSfL | sh]: runc did not terminate sucessfully
------
 > [33/48] RUN curl https://get.wasmer.io -sSfL | sh:
------
failed to solve with frontend dockerfile.v0: failed to build LLB: executor failed running [/bin/sh -c curl https://get.wasmer.io -sSfL | sh]: runc did not terminate sucessfully
script returned exit code 1
```

calavera:
I've removed wasmer in #474. It has never been used in the last year, according to logs, and you can install it with homebrew.

depadiernos:
@ehmicky, were you able to look if the PR that @calavera mentioned can be merged to resolve this issue?

ehmicky:
Yes, I think this should solve the issue.

calavera:
CI should be back to normal now that wasmer is not in the image anymore. Any other issue should be open separately.

Issue #441, opened 2020-06-23
Add Docker image size change to PRs


vbrown608:
When we evaluate PRs to this repo, we want to able to easily answer the question, "how much would this add to the size of our build image?" We could create a github action to add that information to the PR.

verythorough:
Adding to discovery for speedy builds. Also linked in the appropriate section of the [solutions research doc](https://www.notion.so/netlify/Solutions-research-817afa65f3174d8ea58b067e91562ede#fce58d65a1294b08a4b3c0173794d2ba).

vbrown608:
Closed by #516

PR #477, opened 2020-10-01
Add Rust support

### LINKED ISSUE #293: Add Rust to build-image
For using, you know, during builds.

Requested by Pro customer here in our helpdesk.  Others should +1 if they want this too!

- https://app.intercom.io/a/apps/q245f50x/inbox/inbox/1479299/conversations/21364638638

cecton:
Hello @fool 

I just started recently a new OSS project supported by Netlify: https://yewprint.rm.rs (https://github.com/cecton/yewprint) which got approved today  🎉

I'm using Rust and the build time is not short (I don't really mind personally) but I still have build time limit for the sponsored OSS account. So here I am checking if I can help getting Rust support to the Netlify image.

There was already a PR (#320) but it seems to have been closed because of lack of attention. I want to bring this back to life and add wasm-pack and wasm-bindgen support (I suppose everybody who does Rust on Netlify are here to use WASM).

Mandatory gif:

it's-alive-gif

cc @jamen @JeanBarriere @ultrasaurus 

Tasks:

 *  [x] Bring it back to life
 *  [x] Update
 *  [ ] ~Add wasm-pack and wasm-bindgen~

Closes #293 
Related to #320 
Blocked by #476


cecton:
I can't even build the branch xenial locally. There is already an issue #476 

```
\e[1mdownloading: wasmer-linux-amd64
Latest release: 1.0.0-alpha3
Downloading archive from https://github.com/wasmerio/wasmer/releases/download/1.0.0-alpha3/wasmer-linux-amd64.tar.gz
######################################################################## 100.0%
\e[1minstalling: /opt/buildhome/.wasmer
Updating bash profile /opt/buildhome/.profile
we've added the following to your /opt/buildhome/.profile
If you have a different profile please add the following:
\e[2m
# Wasmer
export WASMER_DIR="/opt/buildhome/.wasmer"
[ -s "$WASMER_DIR/wasmer.sh" ] && source "$WASMER_DIR/wasmer.sh"
/opt/buildhome/.wasmer/bin/wasmer: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.27' not found (required by /opt/buildhome/.wasmer/bin/wasmer)
\e[1merror: wasmer was installed, but doesn't seem to be working :(
The command '/bin/sh -c curl https://get.wasmer.io -sSfL | sh' returned a non-zero code: 1
[1] [10:41:40] ~/r/build-image xenial > 
```

calavera:
Hi @cecton. It might be easier for you to wait until we release https://github.com/netlify/build-image/pull/474.

We've tested it already successfully, but we're hosting a conference next week and we want to give ourselves some time to validate it widely. It should not be more than a matter of a few weeks.

cecton:
Hello 👋 

Yes that was my intention! When the issue will be fixed I will complete the PR with wasm-pack and test on my own repo.

Also we might want to fix the version of Rust to avoid the same issue than wasmer.

Thanks a lot for the answer

cecton:
> Also we might want to fix the version of Rust to avoid the same issue than wasmer.

We actually don't need to. rustup ensures rustc is compiled for glibc 2.11+ (64-bit Linux (kernel 2.6.32+, glibc 2.11+)). See https://doc.rust-lang.org/nightly/rustc/platform-support.html

So we can just install rust stable with rustup

run-build-functions.sh
```diff
@@ -645,6 +653,22 @@ install_dependencies() {
     fi
   fi
 
+  # Install rust dependencies and compile the code.
+  if [ -f Cargo.toml ] || [ -f Cargo.lock ]
+  then
+    restore_home_cache ".cargo/registry" "rust deps"
+    check_rust_target_cacheability && restore_home_cache "repo/target" "rust compile output"
+    source $HOME/.cargo/env
+    cargo build --release
```
mraerino: i think this should be specified as the build step by the user instead of being run automatically

run-build-functions.sh
```diff
@@ -675,6 +699,8 @@ cache_artifacts() {
   cache_home_directory ".boot" "boot dependencies"
   cache_home_directory ".composer" "composer dependencies"
   cache_home_directory ".wasmer/cache", "wasmer cache"
+  cache_home_directory ".cargo/registry" "rust deps"
+  check_rust_target_cacheability && cache_home_directory "repo/target" "rust compile output"
```
mraerino: where is `repo/target` coming from?
i think in order to cache something from the user's repository you need to use `cache_cwd_directory`

run-build.sh
```diff
@@ -39,4 +39,6 @@ echo "Executing user command: $cmd"
 eval "$cmd"
 CODE=$?
 
+cache_artifacts
```
mraerino, line 41: why are you adding this?

run-build-functions.sh
cecton: Totally agree! Should I remove the entire `if` block?

run-build-functions.sh
cecton: I don't know I just got this code from a previous PR. `target` is definitely the name of the compilation directory (a bit like `node_modules`).

So should I replace by this?

```
check_rust_target_cacheability && cache_home_directory "$(cache_cwd_directory)"/target "rust compile output"
```

run-build.sh
cecton: I don't know, I'm not the author. Should I remove it?

run-build-functions.sh
```diff
@@ -55,11 +55,19 @@ mkdir -p $NETLIFY_CACHE_DIR/.composer
 mkdir -p $NETLIFY_CACHE_DIR/.gimme_cache/gopath
 mkdir -p $NETLIFY_CACHE_DIR/.gimme_cache/gocache
 mkdir -p $NETLIFY_CACHE_DIR/.wasmer/cache
+mkdir -p $NETLIFY_CACHE_DIR/.cargo/registry
+mkdir -p $NETLIFY_CACHE_DIR/repo/target
```
cecton: This should be replaced by:

```
 mkdir -p $NETLIFY_REPO_DIR/target 
```

run-build-functions.sh
cecton: OH I see!! Nevermind what I said.

```
check_rust_target_cacheability && cache_cwd_directory target "rust compile output"
```

And I will move it up there with the others

run-build.sh
mraerino: yeah, i'd remove it

run-build-functions.sh
mraerino: i think it makes sense to keep the cache handling.
if `cargo` had a command to download dependencies, but not build the project, we could execute that here, since for most languages we seperate the install and build steps.

run-build-functions.sh
mraerino: yup! great

run-build-functions.sh
mraerino: i think it can be removed entirely. `cargo` will create `target` if it doesn't exist.

run-build-functions.sh
cecton: I'm removing this so it will only restore the cache. Fixed in 60fd309

We don't pull the dependencies in Rust as they depend on the build flags.

I don't know what would be better if the dependencies were pulled separately.... let me know if something else should be done here.

run-build-functions.sh
cecton: Fixed by db4ab67

run-build.sh
cecton: Fixed in e4e7354

run-build-functions.sh
cecton: Fixed in 5cd8e76

run-build-functions.sh
```diff
@@ -55,11 +55,18 @@ mkdir -p $NETLIFY_CACHE_DIR/.composer
...
 
 : ${YARN_FLAGS=""}
 : ${NPM_FLAGS=""}
 : ${BUNDLER_FLAGS=""}
 
+check_rust_target_cacheability() {
+  # Check if repo/target is ignored by the repository. If it is, it's OK to
+  # cache it.
+  git --git-dir=${NETLIFY_REPO_DIR}/.git check-ignore -q ${NETLIFY_REPO_DIR}/target
+}
```
cecton: @mraerino Thanks a lot for your help and review btw!! :heart: 

Do you know what is the purpose of this? It seems it checks if `target` is in the `.gitignore` and if that's the case it might not restore `target` from cache.

I think it's good to have a button to purge the cache but here it seems to use a git feature (gitignore) to detect wether or not target should be cached. It doesn't sound right to me and I see no reason why you would do that.

cecton:
@calavera @mraerino this is ready for review! :tada: 

I did not install wasm-pack by default in the end because the tooling depends on the project. Only rustup (the first main common tool to install and manage Rust toolchains) and Rust stable are installed by default.

I added a cache on rustup and `.cargo/bin` so people can change the toolchain easily (they will most likely need wasm) and their tool (wasm-pack, wasm-bindgen, ...) is cached too.

I tested with this command:

```
T=tmp/tmp.RhEPx4A6xs ./test-tools/test-build.sh ~/repos/blueprint-rs 'rustup update; cargo install wasm-pack; ./build.sh'
```

It works like a charm!

wasm-pack would probably be better in a netlify plugin so its directory can be cached also.

(Note: I had to add a line in the test script because the cache was not saved)

run-build-functions.sh
```diff
@@ -57,6 +57,7 @@ mkdir -p $NETLIFY_CACHE_DIR/.composer
 mkdir -p $NETLIFY_CACHE_DIR/.gimme_cache/gopath
 mkdir -p $NETLIFY_CACHE_DIR/.gimme_cache/gocache
 mkdir -p $NETLIFY_CACHE_DIR/.homebrew-cache
+mkdir -p $NETLIFY_CACHE_DIR/.cargo
```
cecton, line 60: This is necessary because I'm caching a few subdirectories to this

run-build.sh
cecton: Ah funny I have re-added this!

I know why the author added this. It's because when you run the test, the cache is restored (by `install_dependencies` in line 29) but it is not saved after the job.

run-build-functions.sh
cecton: In the end I removed it. It makes no sense.

calavera (dismissed):
Thanks a lot for your help @cecton ! I've put this PR in one of our internal backlogs so we can make it available in an upcoming release.

verythorough:
@cecton, thank you so much for taking the time to submit this PR and respond to reviews. As it happens, though, we've simultaneously been working on ways to reduce the size of our build image, and this PR actually increases it by about 20%.

Because build image size impacts the speed of all builds, I don't think we can justify such a significant increase for something that benefits a very small percentage of users, especially now that we have [Homebrew available on the image](https://community.netlify.com/t/now-in-experimental-alpha-homebrew-access-in-the-build-image/26074). With that in mind, I recommend you try using the Homebrew install method, and if you have issues with it, reporting them in issues here or in comments on the [Community announcement thread](https://community.netlify.com/t/now-in-experimental-alpha-homebrew-access-in-the-build-image/26074).

In the meantime, I'm going to close this PR. I'm sorry for that, as I know that communicating our priorities sooner would have saved you time and effort. I don't take that lightly!

cecton:
hello @verythorough 

Thanks I understand. I also think it makes no sense to have such a huge container image if you want to spawn things fast.

The reason why I wanted to add it is because the build time of the image is quite high if I have to install rust and compile my entire project every time. I don't personally mind build time but Netlify counts it and it easily exceeds my OSS account. At the moment I'm have completely disabled builds.

One alternative solution would be to keep at least the caching behavior. This will take no space on your image. If we can keep rustup (a binary, I don't know the size but it should be less than 10Mb), it would be more user friendly.

I think it's worth giving a try, I don't have to do much to adapt my PR for this. Do you agree? (Unless you have other solutions to adapt the cache?)

mraerino:
> One alternative solution would be to keep at least the caching behavior. This will take no space on your image. If we can keep rustup (a binary, I don't know the size but it should be less than 10Mb), it would be more user friendly.

i agree with this proposal. it can greatly improve how we handle Rust. moving most of the toolchain that was previously baked into the image into the cache sounds fine. i'd expect most projects to use a `rust-toolchain` file

calavera:
Homebrew, only installs the compiler, leaving to the user to install any other tooling they need, like cargo, and clippy, which are pretty common dependencies that rustup installs by default and people need to build projects, you can look at the formula's definition:

https://formulae.brew.sh/api/formula/rust.json

I think there is still merit in some of these changes. We can follow the pattern we started with Swift and not install a version by default, Rustup will automatically install one when it detects the `rust-toolchain` file. Rustup is also a small binary that doesn't increase the image size too much.

@cecton can you make the changes to not install stable by default?

The install script should use the option `--default-toolchain none` in https://github.com/netlify/build-image/pull/477/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R493

verythorough:
@cecton Thanks for understanding, and for your suggestion! 

Like @mraerino and @calavera I think it sounds like a worthwhile alternative that does match what we recently arrived at with Swift.

I'd also like to flag @vbrown608, @ehmicky, and @erezrokah, since they're actively working on reducing the build image size right now, and may have some other insights.

cecton:
@verythorough @calavera @mraerino done! :tada: 

I have tested it to be sure and it is still working like a charm. :ok_hand: 

Thank you all for the time and the help. I will write a blog post at some point about how to get started with WASM, Rust and Netlify.

mraerino (dismissed):
i think this looks great!

would you want to test this on a netlify site before we merge? i can internally set this as the build image for any site you want, you'd just have to give me a site name or id

mraerino:
we're shortly going to have a fix for an unrelated bug that currently breaks CI on this.

calavera:
The build should be fixed on master now. Unfortunately, you'll have to rebase this branch to get it green @cecton

mraerino:
or you merge master in which won't require re-approval

cecton:
> i think this looks great!
> 
> would you want to test this on a netlify site before we merge? i can internally set this as the build image for any site you want, you'd just have to give me a site name or id

Sure! Can you do it on yewprint? https://github.com/cecton/yewprint



> The build should be fixed on master now. Unfortunately, you'll have to rebase this branch to get it green @cecton

> or you merge master in which won't require re-approval

I use a merge workflow. It's fixed.


:heavy_check_mark: CI is green! cc @calavera @mraerino

mraerino:
@cecton I updated your site to run with this PR. let us know if it works for you 🙂

cecton:
I tested, works like a charm @mraerino 👌

mraerino:
the team that owns the build process will give this a final look and likely release it next week. in the meantime i'll leave your site configured to use it and will switch it to the regular image once we rolled this.

cecton:
> the team that owns the build process will give this a final look and likely release it next week. in the meantime i'll leave your site configured to use it and will switch it to the regular image once we rolled this.

Thanks! But don't worry about that because I'm stuck with manual deployment for a little longer because of some other issues.  You can change it back now if that makes your life easier @mraerino

vbrown608:
Thanks for your contribution! The changes to the image and install script look good to me. Could you update the `Languages` section of the included software doc? https://github.com/netlify/build-image/blob/xenial/included_software.md#languages. For the version you can say `Not installed by default`. 

(I think Languages makes more sense than Included Software, since we don't list rvm, nvm, etc. there, but open to other ideas on this.)

We're planning to automate this (#441) but in the meantime I'm going to be the bot and note the image size change, which looks reasonable:
```bash
$ docker image inspect cecton-rust:latest --format='{{.Size}}'
4427251610
$ docker image inspect netlify/build:xenial --format='{{.Size}}'
4398783842
```

@rstavchansky I believe we'll also need a change to https://docs.netlify.com/configure-builds/manage-dependencies/#dependency-cache, though I don't think it needs to block merging this PR and it might even be better to run it in production as an alpha for a while before adding it to the docs site. Adding the `needs docs` label.

cecton:
> Could you update the `Languages` section of the included software doc? https://github.com/netlify/build-image/blob/xenial/included_software.md#languages. For the version you can say `Not installed by default`.

Done


> (I think Languages makes more sense than Included Software, since we don't list rvm, nvm, etc. there, but open to other ideas on this.)

Fair enough

vbrown608 (approved):
Looks good :crab: ! We'll include this in the next release of the build image and make a note here when it's live. Thanks again for your work on this.

vbrown608:
This is live in our `Xenial` build image :+1:

mraerino:
@cecton i reset the build image on your site. let me know if there is anything amiss.

davidbarsky:
Hi! I apologize for commenting here, but this change _appears_ to have broken our documentation previews in https://github.com/tokio-rs/tracing/pull/1130. I'm not sure what the right workaround is _or_ what the reporting mechanism y'all prefer me to use, but please let me know!

PR #483, opened 2020-10-21
NETLIFY_INSTALL_JS_DEPENDENCIES skips npm dependency install

### LINKED ISSUE #493: Provide a way to run code before build dependencies are installed.
Similar to https://github.com/netlify/build/issues/1149, as a node user, sometimes, I would need to run some command before `npm install` is run so that my build will not fail. One scenario where this will be useful is described here: https://community.netlify.com/t/common-issue-using-private-npm-modules-on-netlify/795/33

depadiernos:
Setting ~SKIP_NPM_DEPENDENCY_INSTALL~ NETLIFY_INSTALL_JS_DEPENDENCIES to ~true~ false tells the build-image to skip installing npm dependencies (i.e. npm install or yarn install), while still making npm (or yarn if you have NETLIFY_USE_YARN set or have a yarn.lock) available to the build command.

This is non-build-plugin implementation that addresses the issue described here: https://github.com/netlify/build/issues/1149

fixes https://github.com/netlify/build-image/issues/493

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -109,7 +109,7 @@ run_yarn() {
   # The previous pattern doesn't match the end of the string.
   # This removes the flag from the end of the string.
   yarn_local="${yarn_local%--cache-folder *}"
-  if [ "$SKIP_NPM_DEPENDENCY_INSTALL" != "true" ]
+  if [ "$NETLIFY_INSTALL_JS_DEPENDENCIES" != "false" ]
   then
     if yarn install --cache-folder $NETLIFY_BUILD_BASE/.yarn_cache ${yarn_local:+"$yarn_local"}
     then
@@ -149,7 +149,7 @@ run_npm() {
 
   if install_deps package.json $NODE_VERSION $NETLIFY_CACHE_DIR/package-sha
   then
-    if [ "$SKIP_NPM_DEPENDENCY_INSTALL" != "true" ]
+    if [ "$NETLIFY_INSTALL_JS_DEPENDENCIES" != "false" ]
     then
       echo "Installing NPM modules using NPM version $(npm --version)"
       run_npm_set_temp
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -119,7 +119,7 @@ run_yarn() {
       exit 1
     fi
   else
-    echo "Skipping NPM modules install"
+    echo "Skipping Node modules install"
   fi
   export PATH=$(yarn bin):$PATH
 }
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -99,7 +99,7 @@ run_yarn() {
   fi
 
 
-  echo "Installing NPM modules using Yarn version $(yarn --version)"
+  echo "Installing Node modules using Yarn version $(yarn --version)"
   run_npm_set_temp
 
   # Remove the cache-folder flag if the user set any.
@@ -113,7 +113,7 @@ run_yarn() {
   then
     if yarn install --cache-folder $NETLIFY_BUILD_BASE/.yarn_cache ${yarn_local:+"$yarn_local"}
     then
-      echo "NPM modules installed using Yarn"
+      echo "Node modules installed using Yarn"
     else
       echo "Error during Yarn install"
       exit 1
@@ -151,17 +151,17 @@ run_npm() {
   then
     if [ "$NETLIFY_INSTALL_JS_DEPENDENCIES" != "false" ]
     then
-      echo "Installing NPM modules using NPM version $(npm --version)"
+      echo "Installing Node modules using NPM version $(npm --version)"
       run_npm_set_temp
       if npm install ${NPM_FLAGS:+"$NPM_FLAGS"}
       then
-        echo "NPM modules installed"
+        echo "Node modules installed"
       else
-        echo "Error during NPM install"
+        echo "Error during Node install"
         exit 1
       fi
     else
-      echo "Skipping NPM modules install"
+      echo "Skipping Node modules install"
     fi
 
     echo "$(shasum package.json)-$NODE_VERSION" > $NETLIFY_CACHE_DIR/package-sha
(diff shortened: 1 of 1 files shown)


verythorough:
@depadiernos Do you have an issue filed that this PR closes? It can help reviewers understand the customer problems you're trying to solve with this work.

Regarding the name of the variable, we're finding it's helpful to customers if we avoid "double negatives" (i.e. "skip = true" means we don't do the thing).

Instead, I recommend a pattern where the variable names the behavior, and "false" means we don't do it. For example, `INSTALL_NPM_DEPENDENCIES`, which defaults to true. You might also consider replacing the `NPM` part with `NODE` or `JS`, so people don't assume the variable doesn't impact `yarn install`.

cc @rstavchansky in case she has any input on word choices.

depadiernos:
I'm not aware of an issue but will refer to one that is related.

depadiernos:
I'll also rename the env var.

vbrown608:
We've also been moving to prefix environment variables that configure the installation process with `NETLIFY` so they don't collide with any existing variables users may have set - eg. `NETLIFY_INSTALL_NPM_DEPENDENCIES`

rstavchansky:
Keeping in mind Vivian and Jessica's earlier comments, I'd recommend NETLIFY_INSTALL_JS_DEPENDENCIES. This would make it clearer that it could apply to npm or yarn. 

Also, @verythorough it sounds like this may need a corresponding docs issue to update [Netlify configuration variables](https://docs.netlify.com/configure-builds/environment-variables/#netlify-configuration-variables) documentation.

@depadiernos Can you please file a [docs issue](https://github.com/netlify/docs/issues) for this, so it doesn't get lost?

verythorough:
Good call, @rstavchansky! This will warrant a an Updates post in Community, too. @depadiernos will you handle that part as well?

run-build-functions.sh
```diff
@@ -109,13 +109,17 @@ run_yarn() {
...
+    then
+      echo "NPM modules installed using Yarn"
+    else
+      echo "Error during Yarn install"
+      exit 1
+    fi
   else
-    echo "Error during Yarn install"
-    exit 1
+    echo "Skipping NPM modules install"
```
```suggestion
    echo "Skipping Node modules install"
```

ehmicky: Technically those are hosted on npm, even when using Yarn as a client. I am wondering though if some Yarn users might get confused?
What do you think @depadiernos @verythorough?

To keep it consistent with the other messages, I'd recommend still using `NPM modules`, but maybe fixing this in a separate PR (if that change makes sense).

run-build-functions.sh
```diff
@@ -109,13 +109,17 @@ run_yarn() {
...
   # This removes the flag from the end of the string.
   yarn_local="${yarn_local%--cache-folder *}"
-
-  if yarn install --cache-folder $NETLIFY_BUILD_BASE/.yarn_cache ${yarn_local:+"$yarn_local"}
+  if [ "$NETLIFY_INSTALL_JS_DEPENDENCIES" != "false" ]
   then
-    echo "NPM modules installed using Yarn"
+    if yarn install --cache-folder $NETLIFY_BUILD_BASE/.yarn_cache ${yarn_local:+"$yarn_local"}
+    then
+      echo "NPM modules installed using Yarn"
```
ehmicky: See comment below about consistency in messaging here.

run-build-functions.sh
depadiernos: Yea, I've encountered questions about the messaging from yarn users in the past, so I think your concern is valid.

ehmicky: For consistency, we should update the other instances of `NPM modules` as well in this file. Should this be done in a separate PR?

depadiernos: I feel like it's a small enough change. Not too many instances of it, so I'll go ahead and make the change.

depadiernos:
created an issue on Docs here: https://github.com/netlify/docs/issues/776

depadiernos:
created a draft community post here: https://community.netlify.com/t/a-new-build-related-environment-variable-has-been-added/24907 (currently unlisted).

ehmicky (approved):

depadiernos:
@ehmicky, looks like the ci check failed, which seems to be a gpg server error. Can that be ignored?

depadiernos:
re-ran the jenkins test build task. fingers crossed things pass this time.

ehmicky:
This is due to https://github.com/netlify/build-image/issues/476
CI is currently broken.
I think it'd be safer to fix CI first before merging this PR, to ensure CI tests are passing.

rstavchansky:
@depadiernos I came here to see if it's time to do the docs work for this issue, but it looks like I should wait:

> I think it'd be safer to fix CI first before merging this PR, to ensure CI tests are passing.

depadiernos:
> @depadiernos I came here to see if it's time to do the docs work for this issue, but it looks like I should wait:
> 
> > I think it'd be safer to fix CI first before merging this PR, to ensure CI tests are passing.

Yup, we're just waiting for the CI issue to be fixed and we'll be good to go.

ehmicky:
#486 was opened to skip all installations, not just JavaScript.
Do you think this more generic approach might be better? 
I think almost no sites should need to skip only parts of our dependency installation logic, but not others. Having a more generic flag would prevent having to create flags for each specific case.
What do you think?

depadiernos:
I think that PR prevents installing npm or yarn while this still installs those. Just doesn't run the js dependencies (npm install).  It's a bit different scope.

depadiernos:
~need to re-base this now that~ https://github.com/netlify/build-image/pull/474 was merged. Looks like the check is still failing.

verythorough:
Quick update here: as I noted in the related issue (https://github.com/netlify/build-image/pull/486#issuecomment-721253614), we're holding for a bit on these PRs for controlling dependency installation via env vars while we consider a more holistic long-term plan for dependency control. We'll decide what we're doing next week, and report back whether this PR fits with that scheme or not. Thanks for your patience!

erquhart:
@depadiernos can you open an issue describing the problem this PR is solving specifically? Difficult to track this as a standalone PR.

depadiernos:
Got an issue filed. https://github.com/netlify/build-image/issues/493

depadiernos:
Looks like this PR is stale at this point and it looks like there's some planning that would need to happen. In any case, I'm going to go ahead and close it.

PR #484, opened 2020-10-24
Upgrade to the latest Clojure patch version

holyjak:
See https://clojure.org/guides/getting_started#_installation_on_linux. See the [changelog for Clojure](https://clojure.org/releases/tools) and [for tools.deps](https://github.com/clojure/tools.deps.alpha/blob/master/CHANGELOG.md).

Upgrades to the latest  tools.deps.alpha 0.9.821 (was 0.8.584), including support for directly executing clojure functions via `-X`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -375,7 +375,7 @@ RUN mkdir /opt/boot-clj && cd /opt/boot-clj && \
     chmod +x boot && \
     ln -s /opt/boot-clj/boot /usr/local/bin/boot
 
-RUN curl -sL https://download.clojure.org/install/linux-install-1.10.1.492.sh | bash
+RUN curl -sL https://download.clojure.org/install/linux-install-1.10.1.727.sh | bash
 
 USER buildbot
 


mainej:
If this PR and #553 are both merged, we should switch `run-build-functions.sh` to use `clojure -P` instead of `clojure -Spath -Sforce >/dev/null`. The `-P` exec-opt was added in version `1.10.1.672` and more closely matches the intent of the code in `run-build-functions.sh`.

mainej:
@holyjak thanks for opening this PR. I'm still hoping Netlify will merge it, as I'm starting to run into problems building with their old version of Clojure. Would you mind updating the PR to use a [more recent version](https://www.clojure.org/guides/getting_started#_installation_on_linux) of the Clojure CLI (1.10.3.855 as of posting this comment)?

mainej:
I'm giving this another +1. Because the version of Clojure that Netlify uses is outdated, I'm running into problems building with shadow-cljs, version 12.14.5. As of that version, internally shadow-cljs runs `clojure -M -m ...`. The commit that introduced that change https://github.com/thheller/shadow-cljs/commit/c3f738eab4aeae51d2b9ed23abfe5c35cea47275 acknowledged that adding the `-M` option would break users of very outdated Clojure CLIs. I was able to build by downgrading shadow-cljs to 12.14.4, but it would be better to have Netlify support a more recent version of Clojure.

lilactown:
+1 on this, I am also blocked trying to deploy my ClojureScript SPA using shadow-cljs to build + Clojure CLI to manage dependencies

mainej:
@lilactown I've continued to have better luck compiling on Netlify using shadow-cljs version 2.14.4.

Note:
1) It only has to be downgraded in package.json. That is, for now you only need to downgrade the thin NPM package that provides the CLI (`npx shadow-cljs ...`). You can continue to use a more recent Maven package (project.clj, deps.edn, or whatever) which does the real work.
2) When downgrading, your package.json must reference `"shadow-cljs": "2.14.4"`. It's not guaranteed to work if you use, for example, `"^2.14.4"`. I don't understand NPM very well, but I've seen it choose to keep a more recent version in package-lock.json, which is of course what gets used on Netlify.

This workaround doesn't negate the fact that Netlify's Clojure tooling is getting quite out of date. Maybe they're waiting for someone to write a [build plugin](https://docs.netlify.com/configure-builds/build-plugins/create-plugins/) for Clojure. 🤷

brunchboy:
This bit me as well, and confused me a lot when I was trying to deploy an update to my small SPA documenting my shadow-cljs powered [diagramming library for AsciiDoc](https://bytefield-svg.deepsymmetry.org). I found a creaky workaround, but it would be nicer if this just worked the way it should.

austinbirch:
Also caught out by this (thanks for opening the issue for others to see, would have taken me a while to figure out what was going on).

kitop:

With the Homebrew alpha support, it should be able to install this on demand.

mainej:
For those who have been watching this Issue, I wanted to follow up with some details.

I host several ClojureScript projects on Netlify. I use shadow-cljs to compile them. Shadow [dropped support](https://github.com/thheller/shadow-cljs/commit/c3f738eab4aeae51d2b9ed23abfe5c35cea47275) for very old versions of the `clj` command line tools a long time ago. But because Netlify's version of `clj` is very old, I've been stuck using old versions of `shadow-cljs`.

I experimented with @kitop's suggestion—use Linux Homebrew to install a more modern version of `clj`. That worked, and with a few other workarounds I was able to get a recent version of `shadow-cljs` to build my apps on Netlify.

In addition to a newer `clj`, I decided I needed the following things:
* A newer version of `npx`. One of the steps of my build (unrelated to Clojure) needed a newer version of `npx` to work.
* A newer version of `java`. The more recent versions of `shadow-cljs` use a more recent version of the Google Closure compiler, which in turn needs Java 11.

Here are the files I had to add:

**`.nvmrc`** to install a more modern `npm`/`npx`. [Netlify docs](https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript).
```
18
```

**`Brewfile.netlify`** to install a more modern `clj`/`clojure`. [Netlify docs](https://answers.netlify.com/t/now-in-experimental-alpha-homebrew-access-in-the-build-image/26074).
```
# 'brew install'
tap "clojure/tools"
brew "clojure/tools/clojure"
```

Finally I have an executable shell script file in my project. I use this script as the build command in my [`netlify.toml`](https://docs.netlify.com/configure-builds/file-based-configuration/). I added the following code to the top of the script, to download and install java 11 (the temurin JDK for Linux x64 architectures). [temurin docs](https://adoptium.net/installation/archives/). (I tried to install with apt-get, but apt-get requires sudo.)

```sh
wget -q -T 1 -t 1 https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.16.1%2B1/OpenJDK11U-jdk_x64_linux_hotspot_11.0.16.1_1.tar.gz
# Optional; check integrity of download. Would be safer to download the sha256.txt
# file once, commit it to the repo, then run `sha256sum -c` with that on each build.
wget -O- -q -T 1 -t 1 https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.16.1%2B1/OpenJDK11U-jdk_x64_linux_hotspot_11.0.16.1_1.tar.gz.sha256.txt | sha256sum -c

tar xzf OpenJDK11U-jdk_x64_linux_hotspot_11.0.16.1_1.tar.gz

# shadow-cljs needs the jdk on the path. The easiest way to ensure that is to put
# this line in the script file that calls shadow-cljs.
export PATH=$PWD/jdk-11.0.16.1+1/bin:$PATH

java -version
```

So far I'm happy with this solution.
* Installing Clojure via the Brewfile takes about 60 seconds. Not great, but not terrible.
* I expected the JDK download to take several minutes, but it's actually almost instantaneous. I suspect Netlify's network caches that file.
* I'm a little worried about using the Brewfile. It's a 2 year old "alpha" feature with very little documentation. Doesn't give me lots of confidence. But I suppose beggars can't be choosers.

If these steps don't work for you, I'll offer to help but since builds are all so unique, I probably won't have much to add. Otherwise, best of luck!

PR #486, opened 2020-10-31
feat: allow to skip dependencies installation

sileht:
Don't install unnecessary thing for branch containing only static
generated site.

### DIFF
--- a/run-build.sh
+++ b/run-build.sh
@@ -26,14 +26,18 @@ cd $NETLIFY_REPO_DIR
 : ${SWIFT_VERSION="5.2"}
 : ${PYTHON_VERSION="2.7"}
 
-echo "Installing dependencies"
-install_dependencies $NODE_VERSION $RUBY_VERSION $YARN_VERSION $PHP_VERSION $GO_VERSION $SWIFT_VERSION $PYTHON_VERSION
+if [[ -n $SKIP_DEPENDENCIES ]]; then
+  echo "SKIP_DEPENDENCIES set, skipping dependencies installation"
+else
+  echo "Installing dependencies"
+  install_dependencies $NODE_VERSION $RUBY_VERSION $YARN_VERSION $PHP_VERSION $GO_VERSION $SWIFT_VERSION $PYTHON_VERSION
 
-echo "Installing missing commands"
-install_missing_commands
+  echo "Installing missing commands"
+  install_missing_commands
 
-echo "Verify run directory"
-set_go_import_path
+  echo "Verify run directory"
+  set_go_import_path
+fi
 
 echo "Executing user command: $cmd"
 eval "$cmd"


ehmicky:
Hi @sileht,
Thanks for contributing.
If you don't use a build command, dependencies installation should already be skipped. 
Are you in a situation where you need a build command, but do not need to install dependencies?
See also #483.

sileht:
Yes, I have a shell script prepare some stuffs.
Also my repo have a packages.json and a runtime.txt.
This makes netlify install npm and python stuffs while it's not useful at all.

To workaround that, I have a hack:
* create a "netlify/" directory in my repo
* put a hack.sh script inside.
* the hack script contains `cd .. ; ./run_build_script.sh; mv output netlify/output;`
* in netlify, I set the base directory to `netlify/` and the directory to publish to `netlify/output`

Adding this env variable will help me to remove all this craps to inhibit install_dependencies stuffs.

ehmicky:
Thanks for the additional information.
Are you git-committing your `node_modules`? If not, when are your `package.json` dependencies installed?

sileht:
> Thanks for the additional information.
> Are you git-committing your `node_modules`? If not, when are your `package.json` dependencies installed?

No, I don't commit it, and I don't want netlify read my package.json and yarn.lock because they are not part of the website I publish with netlify.

sileht:
#483 can help to not install JS stuff, but it will not fix the python stuff. 

And because I need nothing from the netlify image except it run my build, I propose this simple env var.

ehmicky:
If I understand correctly, you might be in a setup where some parts of your repository are meant for Netlify (the site to deploy), and others are not (the rest of the repository, including a `package.json`). 

You can separate those by using a `base` directory ([documentation](https://docs.netlify.com/configure-builds/common-configurations/#monorepos)). The part meant to be deployed as a site would be grouped into a sub-directory. Then you could point to this sub-directory using the `base` directory setting. Netlify will then only consider that sub-directory during the build, and ignore any other files in that repository. Like this, your repository's `package.json` will not be read and npm dependencies will be skipped. Would this setup work for your use case?

sileht:
Not sure a tool should enforce my repository organization. 

I have another repo where this env will help me a lot, it's for an application that need python 3.8.6 (unsupported by netlify), where I use the same kind of hack to skip `install_dependencies` and then I install the python version I need with pyenv.


If I can do the analogy with heroku, heroku provides buildpacks and users just enable the buildpacks they need.
Heroku just preinstall one buildpack by detecting the language used when you create the site initially after it's up to user to manage the buildpack list.

This makes the experience much better since the initial setup works out of the box and we can customize it as we want, we don't have to wait heroku to support every build use-case, we can even provide our own buildpack or use one from the community if needed.

While I really like what netlify provides, the pain points each time I setup a new netlify site is always this `install_dependencies` I have to fight against because it does many stuffs I don't want or don't do it like I expect it.

That just my 2cents.

ehmicky:
Thanks for the feedback @sileht.
It does seem to me that separating into a sub-directory (or a separate repository) the part of your repository that is a Netlify site from the rest might be a good practice generally speaking. As a platform, it might be reasonable to enforce this separation. This does create a constraint for some repositories to re-organize their files. However, this also allows Netlify to know which files are meant for the Netlify site and which are not. By knowing this, the platform can provide users with several useful optimization and features, automatic installation being one of them.

sileht:
Ok, I don't really want to discuss repo organization, they too many ways to tidy source code with some good and bad reasons.
I agree mine isn't perfect, but I have other tools to deal with.

So, did you plan to accept this (or something like this, like a per language env variables) to let advanced users do the build the way they want and just enjoy the rest of your product where the real value is (at least for me).

Or should I close this and continue with my hack to ensure netlify doesn't setup any languages in my repos ?

verythorough:
Thanks for the PR, @sileht! 

At the moment, we're actually in the process of determining the best long-term method for allowing customers to control automatic dependency installations in their builds. If we add an environment variable, we're making a commitment to continue supporting it forever, so I hesitate to make such a commitment while we're considering other options on the table.

Though expanding the range of build-controlling env vars is one option we're considering, we're also considering ways to allow build plugins to control dependency installation during the [`onPreInstall` event](https://github.com/netlify/build/issues/1149) we're planning to implement in the next few months. I'd love to hear more about how these two different approaches compare in your particular workflow.

We have some other priorities this week, but I expect we'll have decided on a direction by the end of next week. In the meantime, you'll need to keep using your hack a little while longer, but I think it makes sense to keep this PR open in case it fits with the path we end up taking.

sileht:
Having to write a plugin will make things complex while I'm just looking at disabled the automatic dependency installations.

I get why the `automatic dependency installations` is important for the first Netlify onboarding or for a simple/basic build that allows us to quickly enjoy all the other Netlify features (A/B testing, PR preview, analytics, ...)

But I can't see why Netlify image enforces this`automatic dependency installations`; Nelify can't support all existing build tools and all language versions.

And to make users enjoy all the other features, Netlify only needs a directory where static files are generated, whatever how this has been generated.

To rephrase in another way, for me, the`automatic dependency installations` should be an optional feature. Actually, this is a mandatory feature to use to be able to use other ( awesome :wink: ) Netlify features.

k9ert:
I'm having the same issue here and was close to creating a similiar PR. I discussed my case here:
https://answers.netlify.com/t/ignoring-the-requirements-txt/56058

The base-directory might be a workaround for me but i agree with @sileht  that this is not optimal.

> We have some other priorities this week, but I expect we'll have decided on a direction by the end of next week.

Do you have a result out of this?

verythorough:
Sorry we didn't circle back on this before! We did end up changing our plans, and while we won't be implementing the `onPreInstall` event in the near future, we have some other plans in the works.

In the meantime, I can see how the strategy of adding an environment variable would seem like a logical solution, but like I mentioned before, it also adds complexity to the codebase that we can't remove later because people's builds will depend on it, even after better solutions are released. I've also learned that there are other changes that would need to be made to our internal code in order to make a PR like this work, and the team is focused on other priorities right now.

So I'm going to close this PR, as we don't plan to be implementing it at this time. I definitely understand that the strategies of using a `base` directory and/or adding a `cd` command to your build command string aren't always ideal, but they do generally provide a feasible workaround. Thanks for your suggestions and contributions!

PR #489, opened 2020-11-03
Fix homebrew default bundle file

mraerino:
There was a bug where the default for `HOMEBREW_BUNDLE_FILE` was not properly propagated to the environment.


calavera (approved):

ingride (approved):
yay

PR #491, opened 2020-11-05
Remove default Swift version

vbrown608:
Swift is used by few users and accounts for 23% of the size of our base build image. This PR keeps automatic install of Swift during builds, but drops the default version (5.2) from the build image.

With this change, users will need to specify that a build requires Swift in one of three ways:

1. Set the `SWIFT_VERSION` environment variable to the desired version.
2. Include a `.swift-version` file in the build directory.
3. Include a `Package.swift` file in the build directory. If no `.swift-version` file or `SWIFT_VERSION` environment variable is present, we'll default to installing version 5.2.

See https://community.netlify.com/t/upcoming-change-builds-will-no-longer-install-swift-5-2-by-default
Example builds at https://github.com/netlify/buildbot/pull/1055

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -415,9 +415,6 @@ install_dependencies() {
     echo "Attempting Swift version '$SWIFT_VERSION' from .swift-version"
   fi
 
-  swiftenv global ${SWIFT_VERSION} > /dev/null 2>&1
-  export CUSTOM_SWIFT=$?
-
   if [ -d $NETLIFY_CACHE_DIR/swift_version/$SWIFT_VERSION ]
   then
     echo "Started restoring cached Swift version"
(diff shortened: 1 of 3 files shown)


run-build-functions.sh
```diff
@@ -408,33 +407,39 @@ install_dependencies() {
   fi
 
   # Swift Version
-  : ${SWIFT_VERSION="$defaultSwiftVersion"}
+  defaultSwiftVersion="5.2"
```
ehmicky: Should this be on top of the file next to `PIPENV_DEFAULT_PYTHON_VERSION`, and uppercase?

run-build-functions.sh
```diff
@@ -408,33 +407,39 @@ install_dependencies() {
...
-  if [ -d $NETLIFY_CACHE_DIR/swift_version/$SWIFT_VERSION ]
+  # If Package.swift is present and no Swift version is set, use a default
+  if [ -f Package.swift ]
   then
-    echo "Started restoring cached Swift version"
-    rm -rf $SWIFTENV_ROOT/versions/$SWIFT_VERSION
-    cp -p -r $NETLIFY_CACHE_DIR/swift_version/${SWIFT_VERSION} $SWIFTENV_ROOT/versions/
-    swiftenv rehash
-    echo "Finished restoring cached Swift version"
+    : ${SWIFT_VERSION="$defaultSwiftVersion"}
```
ehmicky, line 420: Wouldn't this overwrite the version from `.swift-version` if one is present?

ehmicky: Would it be better to make this a `else if` from the `if [ -f .swift-version ]` block?
This would also remove the need for `: ${...=...}`.

run-build-functions.sh
vbrown608: Yeah, that sounds good. I waffled about because it wasn't exported like the other variables in the block.

run-build-functions.sh
vbrown608: I believe we would still need the conditional override because `SWIFT_VERSION` could be set by the user.

ehmicky: Oh yes, good point.

ehmicky (approved):

verythorough:
Quick note that it would be nice to include the update to `included_software.md` in this change. If not, we can fast-follow in a separate PR. Pinging @rstavchansky for wording.

PR #495, opened 2020-11-12
Update Swift docs in included_software.md

vbrown608:
Updates Swift docs to reflect the changes in #491

### DIFF
--- a/included_software.md
+++ b/included_software.md
@@ -24,9 +24,6 @@ The specific patch versions included will depend on when the image was last buil
   * 7.4
 * Go - `GO_VERSION`
   * 1.12 (default)
-* Swift - `SWIFT_VERSION`, `.swift-version`
-  * 5.2 (default)
-  * Any version that `swiftenv` can install newer than `4.x`. Versions `4.x` and below will not work due to incompatible shared libraries.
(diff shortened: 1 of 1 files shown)


included_software.md
```diff
@@ -35,6 +32,10 @@ The specific patch versions included will depend on when the image was last buil
   * 21 (default)
 * Elixir
   * 1.7 (default)
+* Swift - `SWIFT_VERSION`, `.swift-version`
+  * Not installed by default.
+  * Supports any version that `swiftenv` can install newer than `4.x`. Versions `4.x` and below will not work due to incompatible shared libraries.
+  * 5.2 will be installed if `Package.swift` is present and no version is specified.
```
rstavchansky: [sand] some wording suggestions here:
```suggestion
  * Supports any version that `swiftenv` can install later than `4.x`. Versions `4.x` and earlier will not work due to incompatible shared libraries.
  * 5.2 is installed if `Package.swift` is present and no version is specified with `SWIFT_VERSION` or `.swift-version`.
```

rstavchansky (approved):
Thanks for adding this info. I left some minor editing suggestions.

PR #506, opened 2020-12-12
Build image for Ubuntu 18.04 Bionic Beaver

cecton:
This is not really a PR for xenial, it's a brand new branch.

Hello!

I understand you are working on a new build image based on Ubuntu 20. This is great! I was wondering if in the meantime you would agree to make a new branch bionic for Ubuntu 18.

I simply took xenial and changed it to bionic (fixed a few things here and there). I tested it against my own repo in Rust and it works nicely :ok_hand: but that certainly doesn't cover all of your use cases so **I don't suggest swapping the default to this, I think of it as an opt-in choice for the user**.

Best

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 LABEL maintainer Netlify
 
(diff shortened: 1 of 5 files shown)


kitop:
Closing this since the image for Focal (20.04) was released! 🙌

cecton:
Oh wow great!! :raised_hands:

PR #509, opened 2020-12-21
Disable npm audit in npm install

denis-sokolov:
npm audit is a development tool, here the only thing it does is slow down the build.

npm install by default is a development tool, so it includes features like audit that need to be manually disabled if we do not use npm ci.


kitop:

PR #511, opened 2020-12-21
Keep npm cache

denis-sokolov:
npm keeps its own dependency installation cache which speeds up the install step. It was a notable omission considering we kept caches for other packaging tools.

PR #510, opened 2020-12-21
Fix npm caching issues

denis-sokolov:
npm install is optimized for the development process and if node_modules is in a correct state,
npm install will quit quickly after checking the status of the cache.

Removing the conditional will prevent breakages related to stale cache, like in #113.

After including #509, on a mid-size project with a laptop npm install quits after 1.5s for me.

Also closes #523.

Issue #172, opened 2018-05-13
npm ci support


ekoeryanto:

brycekahle:
Thanks for opening an issue.

We looked into `npm ci` and it would actually be a negative thing for most folks on Netlify. The purpose of the command is intended for CI that is used for testing, rather than deployment, which is what our builds are for. The command re-installs all dependencies, which we've gone to a lot of effort to avoid for performance reasons.

mbrevda:
Hi - I was wondering if you can elaborate a bit?

According to the docs, `npm ci` is

>meant to be used in automated environments such as test platforms, continuous integration, and deployment

So it would seem this can be used for CD as well. Being that npm doesn't need to rebuild the depends graph (as it reads the lockfile) that should speed up the install time. Have your experiences been different? Additionally, `npm ci` guarantees a deterministic, repeatable build - every time, something which cannot be said for `npm i`. 

Further reading:
https://docs.npmjs.com/cli/ci.html
https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci

frangio:
I agree that re-installing all dependencies is undesirable in Netlify, but the benefit of `npm ci` is as @mbrevda mentioned deterministic builds based on the lockfile. I would consider using `npm ci` instead of `npm install` when the `node_modules` directory is not found in the cache, and if there is a lockfile to begin with.

markkasaboski:
The only way `npm install` makes sense to use within an automated build is if we (the developers of the application) edit our package.json files removing all `^, ~, >, >=, <, <= ...` that prefix the versions and get them all to match exactly the version that we want. https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies

Currently, Netlify runs `npm install` very early in the build. This will result in the build process blindly updating any dependencies that are not strictly locked right down to the patch level. We, the application developers, cannot see what has happened. This is dangerous because we, the application developers, will have (hopefully) tested our application against a specific set of dependencies, but now Netlify's build process has altered them, possibly introducing some instability. 

To be able to use `npm install` or `npm ci` should be a nothing more than a simple configuration change within the Netlify UI. This would not break everyone else's builds. It's just an environment (site) level switch that we should have the option to turn on. Or a Team level setting that applies to all environments (sites).

Until then (should it ever happen) this hack will suffice:

[...]

PR #165, opened 2018-03-18
Add an option to skip caching node_modules

jrjohnson:
**Why?**
Some packages do not expect the `node_modules` directory to be cached and
have trouble installing themselves a second time. 

Most recently we have been unable to build a project that contains [sharp](https://www.npmjs.com/package/sharp) which has a compilation step that stores data outside of the `node_modules` directory. We have to manually force a build wit a cleared cache each time.  In the past  we have had similar issues with [node-sass](https://www.npmjs.com/package/node-sass).

**What changed?**
If the `$CACHE_NODE_MODULES` environmental variable is `false` the build will skip
caching or attempting to use a cached `node_modules` directory. If
`$CACHE_NODE_MODULES` is unset or `true` then `node_modules` will continue to
be cached.

PR #513, opened 2020-12-21
Switch to npm ci

### LINKED ISSUE #113: treat package-lock.json the same way we treat package.json if checksum changes
currently we re-run `npm install` in case the checksum of package.json changes:

https://github.com/netlify/build-image/blob/master/run-build-functions.sh#L114

We should do the same if package-lock.json exists and is changed, and someone who knows what they heck they're doing should decide if the presence of package-lock.json means that we don't even worry about changes to package.json at that point while this code gets written

denis-sokolov:
`npm install` is somewhat meant for interactive runs. Here are some reasons to avoid it:
  - it runs a developer-focused audit by default, and while we can disable the audit (#509), there may be more similar features running in the background now or in the future;
  - it is not always meant to be deterministic in a way one would expect from an automated deployment, resulting in unexpected behavior for our users (#172, [forum topic](https://community.netlify.com/t/how-can-i-use-npm-ci-instead-of-npm-install/12570));
  - is difficult to cache correctly, resulting in stale cache and breakages (#113, #165, #510);

`npm ci` is [meant for automated deployments](https://docs.npmjs.com/cli/v7/commands/npm-ci) and does not suffer from the same issues. Instead, it manages its own cache (#511) and ensures a deterministic result.

It still is slower than an npm install under ideal conditions, but it might be less of a concern than previously expected. With the npm cache (#511), an npm ci on a mid-size project on a laptop takes 7-8s for me as opposed to 2-3s for npm install. The command also runs offline, retrieving all packages from the cache.

This PR might be breaking for some user-provided NPM_FLAGS. As npm ci does not take any options, we may be disabling some functionality users expected. An overview of the existing npm install options brings these, probably unlikely, cases:
  - Users with `--dry-run` and package-lock.json will suddenly have their dependencies installed. Assuming they also install their dependencies themselves in a separate step, this will not impact their build success, but will slow it down by adding duplicate effort;
  - Users with `--no-package-lock` will get older versions of dependencies than they expect, likely breaking their builds;
  - Users with `--omit` will get more dependencies than they expect, probably not impacting their build success, but slowing it down;
  - Users with `--ignore-scripts` may have their builds slow down or possibly break.

Closes #113, #114, #165, #172, #509, #510, #511.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -146,20 +146,16 @@ run_npm() {
     fi
   fi
 
-  if install_deps package.json $NODE_VERSION $NETLIFY_CACHE_DIR/package-sha
+  echo "Installing NPM modules using NPM version $(npm --version)"
+  run_npm_set_temp
+  if npm install ${NPM_FLAGS:+"$NPM_FLAGS"}
   then
-    echo "Installing NPM modules using NPM version $(npm --version)"
-    run_npm_set_temp
-    if npm install ${NPM_FLAGS:+"$NPM_FLAGS"}
-    then
-      echo "NPM modules installed"
-    else
-      echo "Error during NPM install"
-      exit 1
-    fi
-
-    echo "$(shasum package.json)-$NODE_VERSION" > $NETLIFY_CACHE_DIR/package-sha
+    echo "NPM modules installed"
+  else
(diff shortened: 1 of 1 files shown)


kitop:

PR #515, opened 2020-12-31
add webp to lfs config

coolya:
Since webp is a modern alternative to the already included jpg, png and gif which is [supported](https://caniuse.com/webp) by all modern browsers, I think it makes sense to include it by default as well.


vbrown608 (approved):
Makes sense! Thanks for the contribution :framed_picture:

PR #516, opened 2021-01-05
Add change in Docker image size to PRs

vbrown608:
Annotates pull requests to this repository with the change in the build image size compared to the one tagged `netlify/build:xenial` on Docker Hub.

This is using the Github Action at https://github.com/vbrown608/whalescale

I think the small change in size on this PR is due to some small change in the dependencies of the build between now and the last build of `xenial`.


ehmicky (approved):
:man_dancing: :woman_dancing:

PR #521, opened 2021-01-12
Fix Rust caching issues

### LINKED ISSUE #505: #477 seems to have broken `cargo doc`-based documentation previews
Hello! The (extremely welcome!) changes in #477 appears to have broken our (relatively niche) use-case where we generate documentation previews for each PR send to `tokio-rs/tracing`. I suspect that with the deployment of #447, the `target` directory that's generated by Cargo is no longer available to `publish` step. This is problematic for us as we use contents of `target/doc` as the deployable documentation preview. In my attempts to debug the deployment failures we're seeing (https://github.com/tokio-rs/tracing/pull/1130), I've developed the following theory:

1. The `cache_cwd_directory` function ([callsite](https://github.com/cecton/build-image/blob/23b30a7e4e3fc25bef908c25b911158b53f4e86d/run-build-functions.sh#L678), [definition](https://github.com/cecton/build-image/blob/23b30a7e4e3fc25bef908c25b911158b53f4e86d/run-build-functions.sh#L755-L757)) moves Cargo's build output to `$NETLIFY_CACHE_DIR`.
2. `cache_cwd_directory` calls [`move_cache`](https://github.com/cecton/build-image/blob/23b30a7e4e3fc25bef908c25b911158b53f4e86d/run-build-functions.sh#L739-L749), which removes the Cargo-generated `target` directory. 

Contemporaneously with https://github.com/netlify/build-image/pull/477#issuecomment-741384078 being posted, we've started noticing the site build errors contain the message "Deploy directory 'target/doc' does not exist". Here's the first build where I've noticed this: https://app.netlify.com/teams/tokio/builds/5fd051070a10d60007567df3.

Anyways, thanks so much for this service! I'm pretty sure this is the first we've ever had to reach out for support!

### LINKED ISSUE #508: Bug: 'target' can't be a deploy directory anymore
I have a project that compiles into the `target` directory, a sub-directory of my repo. My site is configured to use `target` as the 'deploy' directory (aka the 'publish' directory, I think.) This was working until recently. 

Recent changes to this `netlify/build-image` mean that it assumes every project is a rust project, regardless of whether the project uses rust, and that therefore `target` is cache-able. You can see this by `git blame`ing the following line: https://github.com/netlify/build-image/blob/53b83b6bede2920f236b25b6f5a95334320dc849/run-build-functions.sh#L678

After my build finishes (successfully), the caching code `mv`s the `target` directory. See this line: https://github.com/netlify/build-image/blob/53b83b6bede2920f236b25b6f5a95334320dc849/run-build-functions.sh#L746

Because of this, the `/opt/buildhome/repo/target` directory doesn't exist when the deploy starts, and the build fails with the log message "Failed during stage 'building site': Deploy directory 'target' does not exist".

I can think of a few fixes:
1. The build functions could check whether the project uses rust before trying to cache the `target` directory.
2. The build could run the deploy before starting the caching.
3. The caching code could `cp` the cacheable directories, instead of `mv`ing them.
4. The [Support Guide](https://community.netlify.com/t/support-guide-frequently-encountered-problems-during-builds/213) could document forbidden deploy directories.

In the meantime, I'll try to compile to a different directory.

I was able to debug this using the docker image. My prior attempts to fix the problem are documented in [this support ticket](https://community.netlify.com/t/deploy-directory-target-does-not-exist/28413).

calavera:
Fixes https://github.com/netlify/build-image/issues/508
Fixes #505

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -686,7 +686,7 @@ cache_artifacts() {
   cache_home_directory ".composer" "composer dependencies"
   cache_home_directory ".homebrew-cache", "homebrew cache"
   
-  if if [ -f Cargo.toml ] || [ -f Cargo.lock ]
+  if [ -f Cargo.toml ] || [ -f Cargo.lock ]
   then
     cache_home_directory ".rustup" "rust rustup cache"
     cache_home_directory ".cargo/registry" "rust cargo registry cache"
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -675,7 +675,11 @@ cache_artifacts() {
   cache_cwd_directory ".venv" "python virtualenv"
   cache_cwd_directory ".build" "swift build"
   cache_cwd_directory ".netlify/plugins" "build plugins"
-  cache_cwd_directory "target" "rust compile output"
+
+  if [ -f Cargo.toml ] || [ -f Cargo.lock ]
+  then
+    cache_cwd_directory "target" "rust compile output"
+  fi
 
   cache_home_directory ".yarn_cache" "yarn cache"
   cache_home_directory ".cache/pip" "pip cache"
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -689,10 +689,10 @@ cache_artifacts() {
   cache_home_directory ".boot" "boot dependencies"
   cache_home_directory ".composer" "composer dependencies"
   cache_home_directory ".homebrew-cache", "homebrew cache"
+  cache_home_directory ".rustup" "rust rustup cache"
   
   if [ -f Cargo.toml ] || [ -f Cargo.lock ]
   then
-    cache_home_directory ".rustup" "rust rustup cache"
     cache_home_directory ".cargo/registry" "rust cargo registry cache"
     cache_home_directory ".cargo/bin" "rust cargo bin cache"
   fi
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -685,9 +685,13 @@ cache_artifacts() {
   cache_home_directory ".boot" "boot dependencies"
   cache_home_directory ".composer" "composer dependencies"
   cache_home_directory ".homebrew-cache", "homebrew cache"
-  cache_home_directory ".rustup" "rust rustup cache"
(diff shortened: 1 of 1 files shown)


run-build-functions.sh
```diff
@@ -685,9 +685,13 @@ cache_artifacts() {
   cache_home_directory ".boot" "boot dependencies"
   cache_home_directory ".composer" "composer dependencies"
   cache_home_directory ".homebrew-cache", "homebrew cache"
-  cache_home_directory ".rustup" "rust rustup cache"
-  cache_home_directory ".cargo/registry" "rust cargo registry cache"
-  cache_home_directory ".cargo/bin" "rust cargo bin cache"
+  
+  if if [ -f Cargo.toml ] || [ -f Cargo.lock ]
```
vbrown608:
```suggestion
  if [ -f Cargo.toml ] || [ -f Cargo.lock ]
```

calavera:
I've opted for changing only the behavior of how we cache the `target` directory, instead of all cached directories to minimize the impact of these changes.

calavera:
@vbrown608 I've tested this with an internal build and it seems to work as expected

run-build-functions.sh
```diff
@@ -685,9 +689,13 @@ cache_artifacts() {
   cache_home_directory ".boot" "boot dependencies"
   cache_home_directory ".composer" "composer dependencies"
   cache_home_directory ".homebrew-cache", "homebrew cache"
-  cache_home_directory ".rustup" "rust rustup cache"
-  cache_home_directory ".cargo/registry" "rust cargo registry cache"
-  cache_home_directory ".cargo/bin" "rust cargo bin cache"
+  
+  if [ -f Cargo.toml ] || [ -f Cargo.lock ]
+  then
+    cache_home_directory ".rustup" "rust rustup cache"
```
mraerino: I think we should try to cache this regardless of whether there are Cargo config files

calavera: updated

mraerino (approved):
awesome!

vbrown608 (approved):

Issue #479, opened 2020-10-07
Support for caching yarn workspaces' node_modules


kamsar:
**Which problem is this feature request solving?**

https://community.netlify.com/t/gatsby-js-yarn-workspaces-and-caching-node-modules/6571

Given a structure such as this with Yarn Workspaces:

```
package.json - monorepo package defines yarn workspaces
yarn.lock - this root yarn.lock has ALL the dependencies in it for the whole set of workspace packages
node_modules - contains ALL modules for the workspace that are not specific to a single package (module hoisting)
packages/
    package1/
        {other package files including src}
        package.json - a monorepo package that we are deploying to Netlify
        yarn.lock - a fake empty lockfile so Netlify Build uses yarn (not actually used by yarn at all)
        netlify.toml - settings for this deployment
        node_modules - contains symlinks and packages specific to `package1` only (i.e. eslint would be in /node_modules)
```

We need to set our base directory to `packages/package1` so that the netlify.toml is read correctly. However doing this means `/node_modules` is NOT cached, only `/packages/package1/node_modules` which is practically empty. This means every build is adding ~120 sec to re-download the root modules from npm.

**Describe the solution you'd like**

* Detect yarn workspaces in use (presence of `workspaces` key in some ancestor `package.json` would indicate this)
* Automatically use yarn instead of `npm` without needing the fake `yarn.lock` in the package folder (nice to have)
* Cache both the workspace and local `node_modules` folders when a workspace is being used

A more manual solution might be to allow specifying paths to cache explicitly, or setting an environment variable.

**Describe alternatives you've considered**

Build plugins let me cache arbitrary things but unfortunately:

* They currently run after dependencies have been installed, so they are too late to expand the explicitly cached modules before buildbot has already done a full install
* Build plugin cache explicitly refuses to cache dotfiles (hidden files), and the yarn cache is corrupt without these so even with a build plugin yarn is unable to successfully restore cache

**Can you submit a pull request?**

Possibly with some suggestions as to desired approach.

ehmicky:
Hi @kamsar,

Thanks for filing this issue.
This is a feature we are currently missing.
This has been already reported in #399, #196, #432 and #433. Also related to #465.

plum-dawg:
Has there been any movement on this issue?

kaganjd:
[...]

Issue #432, opened 2020-05-22
Yarn workspaces are not properly cached


ehmicky:
Related: #399, #196.

When using Yarn workspaces, each new build is running `yarn` again. It seems that `node_modules` are not properly cached. 

This behavior is unrelated the Netlify Build beta (it happens both with and without it).

Repository to reproduce:

`package.json`

```json
{
  "private": true,
  "workspaces": ["site"]
}
```

`netlify.toml`

```toml
[build]
command = "node --version"
```

`site/package.json`

```json
{
  "name": "site",
  "version": "1.0.0",
  "license": "MIT",
  "dependencies": {
    "gatsby": "^2.4.2"
  }
}
```

Run `yarn` to create a `yarn.lock`. Then each deploy keeps showing:

```
6:51:24 PM: Started restoring cached node modules
6:51:24 PM: Finished restoring cached node modules
6:51:24 PM: Started restoring cached yarn cache
6:51:24 PM: Finished restoring cached yarn cache
6:51:25 PM: Installing NPM modules using Yarn version 1.17.0
6:51:25 PM: yarn install v1.17.0
6:51:25 PM: [1/4] Resolving packages...
6:51:26 PM: [2/4] Fetching packages...
6:51:51 PM: info fsevents@2.1.3: The platform "linux" is incompatible with this module.
6:51:51 PM: info "fsevents@2.1.3" is an optional dependency and failed compatibility check. Excluding it from installation.
6:51:51 PM: info fsevents@1.2.13: The platform "linux" is incompatible with this module.
6:51:51 PM: info "fsevents@1.2.13" is an optional dependency and failed compatibility check. Excluding it from installation.
6:51:51 PM: [3/4] Linking dependencies...
6:51:51 PM: warning "workspace-aggregator-bb3d6533-7bc7-4a3e-8af3-ed67450a57b9 > site > gatsby@2.22.9" has unmet peer dependency "react@^16.4.2".
6:51:51 PM: warning "workspace-aggregator-bb3d6533-7bc7-4a3e-8af3-ed67450a57b9 > site > gatsby@2.22.9" has unmet peer dependency "react-dom@^16.4.2".
6:51:51 PM: warning "workspace-aggregator-bb3d6533-7bc7-4a3e-8af3-ed67450a57b9 > site > gatsby > @pmmmwh/react-refresh-webpack-plugin@0.3.2" has incorrect peer dependency "react-refresh@^0.8.2".
[...]

Issue #399, opened 2020-04-27
Yarn workspaces support improvements


ehmicky:
For monorepos where each package has its own Netlify Site, we recommend using a different base directory for each of them, and optionally a different `netlify.toml`. At least that's my assumption, please correct me if I'm wrong.

The problem is that, when using Yarn workspaces, the `yarn.lock` file will be at the top-level, not inside each package. Because of this, we currently don't detect it and we use `npm` instead of `yarn`.

One workaround is to create an empty `yarn.lock` file inside packages directories, but that's hacky.

This problem was reported by @50bbx [here](https://github.com/jlengstorf/netlify-plugin-gatsby-cache/issues/17#issuecomment-620033951).

pixelbreaker:
I'm having really slow builds due to this issue. I have a large app in `/www` and related common UI in `/components`. When I set up as a monorepo with a base folder of `/www` in the Netlify UI, and hacked a `yarn.lock` into `/www` with a symlink to the one in the root, then Yarn runs and redownloads/compiles all packages every time a  build runs. 

This is costing us about 95 seconds of build minutes for every single deploy.

slorber:
Hey

Having an empty yarn.lock in a sub package seems to work to tell netlify to use Yarn, however I encouter the same issue as @pixelbreaker with Yarn downloading each time dependencies.

As dependencies end up being in both root/node_modules + root/packages/mySiteToDeploy/node_modules, it's not totally clear how Netlify handles caching


Also, there is a "yarn install --focus" command that we should be able to leverage so that Yarn download only the dependencies for the package we are trying to build. Currently if you have a very large monorepo with a packages having very different dependencies, all dependencies have to be downloaded each time. (see https://classic.yarnpkg.com/blog/2018/05/18/focused-workspaces/)


Also, on my personal website ([opensource](https://github.com/slorber/sebastienlorber.com) gatsby site, no workspace) I had `yarn install && yarn build` as a build command. I was surprised that it would redownload dependencies on each commit. As adviced by someone from Forestry who asked someone of the Netlify team, moving to just `yarn build` did solve the problem (as Netlify download deps anyway), but it's not totally clear to me why? 

For monorepo support doesn't it make sense to let the user add himself `yarn install` or `yarn install --focus` to his build command? Both cases make sense to me, it depends on if the deployed app first need to build extra packages of the monorepo (in which case having all deps seems necessary) or if the extra packages from which the app depends are already build (in which case we can use --focus to make the build faster)

[...]

PR #526, opened 2021-02-22
feat: better cache support for JS workspaces

JGAntunes:
Based on the discussion started in https://github.com/netlify/pod-workflow/issues/139 (which is in turn based on a lot of the open PRs and issues in this repo), this PR adds support for [js workspaces](https://classic.yarnpkg.com/en/docs/workspaces/).

Despite being a feature originally supported by `yarn` only, `npm v7` is [catching up](https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/), with on-going discussions and RFCs (see https://github.com/npm/rfcs/pull/117 and https://github.com/npm/rfcs/pull/273) to add further functionality. These however still work differently, `npm` is unable to detect it is within a project with multiple workspaces if we `cd` into a sub-directory. For `netlify` (if we use the [`base` configuration property](https://docs.netlify.com/configure-builds/file-based-configuration/#sample-file)) our caching strategy still holds up, since we just install and cache a workspace package as it was a regular js project. For `yarn` however, installs in sub directories are able to infer they're part of a `workspace`, meaning dependencies will be hoisted and all the sibling packages will be installed also. This means that, in order to achieve the same caching experience we have for other JS projects, we need to cache the root `node_modules` as well as all the workspace `node_modules`.

This solution works for both `yarn v1` and `yarn v2` (that [rely on `node_modules` `nodeLinker` config](https://yarnpkg.com/configuration/yarnrc#nodeLinker)).

I've put in place an env variable - `NETLIFY_YARN_WORKSPACES` - that we can rely on to use as feature flag by hooking this up in the `buildbot` side with launch darkly.

Addresses #399 #432 #196 #479 #196 

## Tests

I've run this locally on a monorepo using:
- [yarn](https://github.com/JGAntunes/gatsby-yarn-netlify-test/tree/multiple-netlify-sites-yarn)
- [yarn v2](https://github.com/JGAntunes/gatsby-yarn-netlify-test/tree/multiple-netlify-sites-yarn-v2)
- [npm](https://github.com/JGAntunes/gatsby-yarn-netlify-test/tree/multiple-netlify-sites)

An example output of a cached execution with `yarn`:
```sh
Finished restoring cached yarn cache
NETLIFY_YARN_WORKSPACES feature flag set
yarn workspaces v1.22.4
{
  "gatsby-starter-blog-1": {
    "location": "packages/blog-1",
    "workspaceDependencies": [],
    "mismatchedWorkspaceDependencies": []
  },
  "gatsby-starter-blog-2": {
    "location": "packages/blog-2",
    "workspaceDependencies": [],
    "mismatchedWorkspaceDependencies": []
  }
}
Done in 0.12s.
Started restoring workspace packages/blog-1 node modules
Finished restoring workspace packages/blog-1 node modules
Started restoring workspace packages/blog-2 node modules
Finished restoring workspace packages/blog-2 node modules
Started restoring workspace root node modules
Finished restoring workspace root node modules
Installing NPM modules using Yarn version 1.22.4
mkdir: cannot create directory ‘/opt/buildhome/tmp’: File exists
yarn install v1.22.4
[1/4] Resolving packages...
success Already up-to-date.
Done in 1.25s.
NPM modules installed using Yarn

(...)

Caching artifacts
Started saving workspace packages/blog-1 node modules
Finished saving workspace packages/blog-1 node modules
Started saving workspace packages/blog-2 node modules
Finished saving workspace packages/blog-2 node modules
Started saving workspace root node modules
Finished saving workspace root node modules
Started saving build plugins
Finished saving build plugins
Started saving yarn cache
Finished saving yarn cache
Started saving go dependencies
Finished saving go dependencies
```

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -40,6 +40,7 @@ mkdir -p $NETLIFY_CACHE_DIR/swift_version
 
 # pwd caches
 mkdir -p $NETLIFY_CACHE_DIR/node_modules
+mkdir -p $NETLIFY_CACHE_DIR/js-workspaces
 mkdir -p $NETLIFY_CACHE_DIR/.bundle
 mkdir -p $NETLIFY_CACHE_DIR/bower_components
 mkdir -p $NETLIFY_CACHE_DIR/.venv
@@ -102,6 +103,32 @@ run_yarn() {
   fi
 
 
+  if [ "$NETLIFY_YARN_WORKSPACES" = "true" ]
+  then
+    echo "NETLIFY_YARN_WORKSPACES feature flag set"
+    # Ignore path env var required in order to support yarn 2 projects
+    if YARN_IGNORE_PATH=1 yarn workspaces info
+    then
+      # Extract all the packages and respective locations. .data will be a JSON object like
+      # {
+      #   "my-package-1": {
+      #     "location": "packages/blog-1",
+      #     "workspaceDependencies": [],
+      #     "mismatchedWorkspaceDependencies": []
+      #   },
+      #   (...)
+      # }
+      # We need to cache all the node_module dirs, or we'll always be installing them on each run
+      local package_locations="$(YARN_IGNORE_PATH=1 yarn --json workspaces info | jq -r '.data | fromjson | to_entries | .[].value.location | @sh'| tr -d \')"
+      restore_js_workspace_cache "$package_locations"
+    else
+      echo "No workspace detected"
+      restore_cwd_cache node_modules "node modules"
+    fi
+  else
+    restore_cwd_cache node_modules "node modules"
+  fi
+
   echo "Installing NPM modules using Yarn version $(yarn --version)"
   run_npm_set_temp
 
@@ -130,6 +157,8 @@ run_npm_set_temp() {
 }
 
 run_npm() {
+  restore_cwd_cache node_modules "node modules"
+
   if [ -n "$NPM_VERSION" ]
   then
     if [ "$(npm --version)" != "$NPM_VERSION" ]
@@ -475,7 +504,6 @@ install_dependencies() {
 
   if [ -f package.json ]
   then
-    restore_cwd_cache node_modules "node modules"
     if [ "$NETLIFY_USE_YARN" = "true" ] || ([ "$NETLIFY_USE_YARN" != "false" ] && [ -f yarn.lock ])
     then
       run_yarn $YARN_VERSION
@@ -676,6 +704,11 @@ cache_artifacts() {
   cache_cwd_directory ".build" "swift build"
   cache_cwd_directory ".netlify/plugins" "build plugins"
 
+  if [ "$NETLIFY_YARN_WORKSPACES" == "true" ] && [ -f package.json ]
+  then
+    cache_js_workspaces
+  fi
+
   if [ -f Cargo.toml ] || [ -f Cargo.lock ]
   then
     cache_cwd_directory_fast_copy "target" "rust compile output"
@@ -690,7 +723,7 @@ cache_artifacts() {
   cache_home_directory ".composer" "composer dependencies"
   cache_home_directory ".homebrew-cache", "homebrew cache"
   cache_home_directory ".rustup" "rust rustup cache"
-  
+
   if [ -f Cargo.toml ] || [ -f Cargo.lock ]
   then
     cache_home_directory ".cargo/registry" "rust cargo registry cache"
@@ -777,6 +810,43 @@ cache_home_directory() {
 
 restore_cwd_cache() {
   move_cache "$NETLIFY_CACHE_DIR/$1" "$PWD/$1" "restoring cached $2"
+}
+
+#
+# Restores node_modules dirs cached for js workspaces
+# See https://github.com/netlify/pod-workflow/issues/139/ for more context
+#
+# Expects:
+# $1 array containing package locations relative to the repo's root
+restore_js_workspaces_cache() {
+  local locations=$1
+  local cache_dir="$NETLIFY_CACHE_DIR/js-workspaces"
+
+  # Retrieve each workspace node_modules
+  for location in "${locations[@]}"; do
+    move_cache "$cache_dir/$location/node_modules" "$NETLIFY_REPO_DIR/$location/node_modules" "restoring workspace $location node modules"
+  done
+  # Retrieve hoisted node_modules
+  move_cache "$cache_dir/node_modules" "$NETLIFY_REPO_DIR/node_modules" "restoring root node modules"
+  # Keep a record of the workspaces in the project in order to cache them later
+  echo "${locations[@]}" > "$cache_dir/.workspace_locations"
+}
+
+#
+# Caches node_modules dirs from js workspaces. It acts based on the presence of a
+# `.workspace_locations` file previously generated in `restore_js_workspaces_cache()`
+#
+cache_js_workspaces() {
+  local cache_dir="$NETLIFY_CACHE_DIR/js-workspaces"
+  if [ -f "$cache_dir/.workspace_locations" ]
+  then
+    local locations=($(cat "$cache_dir/.workspace_locations"))
+    for location in "${locations[@]}"; do
+      move_cache "$NETLIFY_REPO_DIR/$location/node_modules" "$cache_dir/$location/node_modules" "saving workspace $location node modules"
+    done
+    # Retrieve hoisted node_modules
+    move_cache "$NETLIFY_REPO_DIR/node_modules" "$cache_dir/node_modules" "saving root node modules"
+  fi
 }
 
 cache_cwd_directory() {
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -108,9 +108,11 @@ run_yarn() {
   if [ "$NETLIFY_YARN_WORKSPACES" = "true" ]
   then
     echo "NETLIFY_YARN_WORKSPACES feature flag set"
+    local workspace_output
     # YARN_IGNORE_PATH will ignore the presence of a local yarn executable (i.e. yarn 2) and default
     # to using the global one (which, for now, is alwasy yarn 1.x). See https://yarnpkg.com/configuration/yarnrc#ignorePath
-    if YARN_IGNORE_PATH=1 yarn workspaces info
+    workspace_output=$(YARN_IGNORE_PATH=1 yarn workspaces info)
+    if $?
     then
       local package_locations
       # Extract all the packages and respective locations. .data will be a JSON object like
@@ -123,7 +125,7 @@ run_yarn() {
       #   (...)
       # }
       # We need to cache all the node_module dirs, or we'll always be installing them on each run
-      mapfile -t package_locations <<< "$(YARN_IGNORE_PATH=1 yarn --json workspaces info | jq -r '.data | fromjson | to_entries | .[].value.location')"
+      mapfile -t package_locations <<< "$workspace_output"
       restore_js_workspaces_cache "${package_locations[@]}"
     else
       echo "No workspace detected"
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -109,10 +109,12 @@ run_yarn() {
   then
     echo "NETLIFY_YARN_WORKSPACES feature flag set"
     local workspace_output
+    local workspace_exit_code
     # YARN_IGNORE_PATH will ignore the presence of a local yarn executable (i.e. yarn 2) and default
     # to using the global one (which, for now, is alwasy yarn 1.x). See https://yarnpkg.com/configuration/yarnrc#ignorePath
-    workspace_output=$(YARN_IGNORE_PATH=1 yarn workspaces info)
-    if $?
+    workspace_output="$(YARN_IGNORE_PATH=1 yarn workspaces --json info )"
+    workspace_exit_code=$?
+    if [ $workspace_exit_code -eq 0 ]
     then
       local package_locations
       # Extract all the packages and respective locations. .data will be a JSON object like
@@ -125,7 +127,7 @@ run_yarn() {
       #   (...)
       # }
       # We need to cache all the node_module dirs, or we'll always be installing them on each run
-      mapfile -t package_locations <<< "$workspace_output"
+      mapfile -t package_locations <<< "$(echo "$workspace_output" | jq -r '.data | fromjson | to_entries | .[].value.location')"
       restore_js_workspaces_cache "${package_locations[@]}"
     else
       echo "No workspace detected"
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -824,17 +824,17 @@ restore_cwd_cache() {
 # Expects:
 # $@ each argument should be a package location relative to the repo's root
 restore_js_workspaces_cache() {
-  local locations=("$@")
+  # Keep a record of the workspaces in the project in order to cache them later
+  NETLIFY_JS_WORKSPACE_LOCATIONS=("$@")
+
   local cache_dir="$NETLIFY_CACHE_DIR/js-workspaces"
 
   # Retrieve each workspace node_modules
-  for location in "${locations[@]}"; do
+  for location in "${NETLIFY_JS_WORKSPACE_LOCATIONS[@]}"; do
     move_cache "$cache_dir/$location/node_modules" "$NETLIFY_REPO_DIR/$location/node_modules" "restoring workspace $location node modules"
   done
   # Retrieve hoisted node_modules
   move_cache "$cache_dir/node_modules" "$NETLIFY_REPO_DIR/node_modules" "restoring workspace root node modules"
-  # Keep a record of the workspaces in the project in order to cache them later
-  NETLIFY_JS_WORKSPACE_LOCATIONS=$(printf '%s\n' "${locations[@]}")
 }
 
 #
@@ -842,7 +842,8 @@ restore_js_workspaces_cache() {
 # via the `NETLIFY_JS_WORKSPACE_LOCATIONS` variable, or looks at the node_modules in the cwd.
 #
 cache_node_modules() {
-  if [ -z "$NETLIFY_JS_WORKSPACE_LOCATIONS" ]
+  # Check the number of workspace locations detected
+  if [ "${#NETLIFY_JS_WORKSPACE_LOCATIONS[@]}" -eq 0 ]
   then
     cache_cwd_directory "node_modules" "node modules"
   else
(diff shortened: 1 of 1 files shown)


run-build-functions.sh
```diff
@@ -102,6 +103,32 @@ run_yarn() {
   fi
 
 
+  if [ "$NETLIFY_YARN_WORKSPACES" = "true" ]
+  then
+    echo "NETLIFY_YARN_WORKSPACES feature flag set"
+    # Ignore path env var required in order to support yarn 2 projects
+    if YARN_IGNORE_PATH=1 yarn workspaces info
```
erezrokah: [sand] Can we clarify a bit what `YARN_IGNORE_PATH` does?

erezrokah (dismissed):
🚀

erezrokah:
> (that [rely on `node_modules` `nodeLinker` config](https://yarnpkg.com/configuration/yarnrc#nodeLinker)).

Do users need to manually configure that?

JGAntunes:
> > (that [rely on `node_modules` `nodeLinker` config](https://yarnpkg.com/configuration/yarnrc#nodeLinker)).
> 
> Do users need to manually configure that?

@erezrokah yes, but the purpose of this option is for users who are using `berry` (yarn v2) and want to opt-out of the [default `pnp` strategy](https://next.yarnpkg.com/features/pnp) - or at least that's my understanding of it. For projects relying on the default yarn `pnp` strategy I believe there are further considerations and changes that need to happen to our build-image to be able to effectively cache those projects - https://github.com/netlify/build-image/pull/465/files#diff-6ad81dc23c4c92a6e6193db3b05e306293aa75ee25d48ae952c6259c601e8b23R117 - However I need to better look into it 👍

run-build-functions.sh
JGAntunes: Tried to had a better description. Let me know what you think @erezrokah 👍 The naming for the variable vs its final purpose can be a bit confusing 😅

run-build-functions.sh
```diff
@@ -779,6 +820,54 @@ restore_cwd_cache() {
...
+  local cache_dir="$NETLIFY_CACHE_DIR/js-workspaces"
+
+  # Retrieve each workspace node_modules
+  for location in "${locations[@]}"; do
+    move_cache "$cache_dir/$location/node_modules" "$NETLIFY_REPO_DIR/$location/node_modules" "restoring workspace $location node modules"
+  done
+  # Retrieve hoisted node_modules
+  move_cache "$cache_dir/node_modules" "$NETLIFY_REPO_DIR/node_modules" "restoring workspace root node modules"
+  # Keep a record of the workspaces in the project in order to cache them later
+  echo "${locations[@]}" > "$NETLIFY_TMP_DIR/.workspace_locations"
```
JGAntunes: Because we need to have the ability to cache each workspace location at the end of the run we use the `restore_js_workspaces_cache` function as a way to set a temporary file, containing the package locations, that we can later use when caching directories.

run-build-functions.sh
```diff
@@ -669,9 +702,17 @@ install_dependencies() {
...
   cache_cwd_directory "bower_components" "bower components"
-  cache_cwd_directory "node_modules" "node modules"
+
+  if [ "$NETLIFY_YARN_WORKSPACES" == "true" ]
+  then
+    cache_node_modules
+  else
+    cache_cwd_directory "node_modules" "node modules"
+  fi
+
```
JGAntunes, line 718: The idea is to, once we can drop the feature flag, we keep only the `cache_node_modules` function containing all the logic relative to `node_modules` caching.

run-build-functions.sh
```diff
@@ -102,6 +107,33 @@ run_yarn() {
...
+      # {
+      #   "my-package-1": {
+      #     "location": "packages/blog-1",
+      #     "workspaceDependencies": [],
+      #     "mismatchedWorkspaceDependencies": []
+      #   },
+      #   (...)
+      # }
+      # We need to cache all the node_module dirs, or we'll always be installing them on each run
+      local package_locations=($(YARN_IGNORE_PATH=1 yarn --json workspaces info | jq -r '.data | fromjson | to_entries | .[].value.location | @sh'| tr -d \'))
```
JGAntunes: The idea here is to rely on the `yarn workspaces info --json` to give us the detailed list of locations where packages are located. That way we can cache all the relevant `node_modules` without needing to guess and run random searches.

run-build-functions.sh
```diff
@@ -59,6 +60,10 @@ mkdir -p $NETLIFY_CACHE_DIR/.gimme_cache/gocache
 mkdir -p $NETLIFY_CACHE_DIR/.homebrew-cache
 mkdir -p $NETLIFY_CACHE_DIR/.cargo
 
+
+# tmp dir
+NETLIFY_TMP_DIR=$(mktemp -d)
```
JGAntunes: Needed a `tmp` dir to be used for the `.workspace_locations` file and figured I should probably add a generic dir to be used for other purposes if needed. Let me know what you all think or if there's any concerns with this. CC @vbrown608

run-build-functions.sh
verythorough: @JGAntunes Are these env vars intended to be user-editable in the long term? If so, we'll need to document them. (cc @rstavchansky)

Is there anything else that should be documented? Or is this automatic behavior that's considered more "under the hood"?

run-build-functions.sh
verythorough: 🤔 I wonder if it would be useful to implement a similar check in the new site flow and/or framework-info library. I'm not sure if that's feasible, but it could be helpful for monorepo site setup.

run-build-functions.sh
erezrokah: Thanks @JGAntunes - so we are forcing `yarn@1` just for the workspaces command? 
And this works since the configuration hasn't changed between v1 and v2?

run-build-functions.sh
ehmicky: Would a variable work instead of a file?
This would remove the need for `NETLIFY_TMP_DIR` and would be faster.

run-build-functions.sh
ehmicky: Should we handle file paths with spaces in them, or is this too much of an edge case? 
I do not believe this currently works for them, due to `$()` not being quoted. However quoting it would not work with array creation `( )`.

run-build-functions.sh
JGAntunes: > Are these env vars intended to be user-editable in the long term?
> Is there anything else that should be documented? Or is this automatic behavior that's considered more "under the hood"?

The only new user-editable env var we're introducing is `NETLIFY_YARN_WORKSPACES`, and it shouldn't be editable in the long term, hence no need to document it IMO CC @rstavchansky. Theoretically speaking, users could set an environment variable `NETLIFY_YARN_WORKSPACES=true` that would trigger this behaviour (instead of relying on a feature flag on our side to do it - which is our plan to roll this out) but I don't really see that as major concern. The plan is for this behaviour to eventually become default without the need of any env vars (we infer the presence of a `workspace` monorepo on our end).

That being said, later on (maybe once this has been rolled out accordingly), I think it would be a good idea to produce some examples of monorepo usages with Netlify as it is not a trivial setup to do and, to my understanding, we don't have a write up on this (correct me if I'm wrong though). I think there's a lot of workarounds that originated in the issues above that are no longer applicable, so documenting that could probably be beneficial. Happy to move this discussion back to - https://github.com/netlify/pod-workflow/issues/139 - if you prefer 👍 CC @rstavchansky @verythorough

JGAntunes: > Thanks @JGAntunes - so we are forcing yarn@1 just for the workspaces command?
And this works since the configuration hasn't changed between v1 and v2?

Exactly @erezrokah! The configuration is still extracted from the `workspaces` entry in the `package.json` and that hasn't changed (from what I managed to look into), that way we are able to validate the presence of `workspaces` without the need of a separate branch for the `yarn berry` case.

Is that clear enough from the comment and usage? I'm happy to provide further information in there if you think like it makes sense 👍

run-build-functions.sh
JGAntunes: I was about to write that it should be relatively simple to achieve it since we could rely directly in the `package.json`, but completely forgot that the `workspaces` entry will [live in the root `package.json`](https://classic.yarnpkg.com/en/docs/workspaces/) which I guess won't be accessible out of the box for the `framework-info`... 😐

We can maybe open an issue in the repo and discuss it over there? CC @verythorough

JGAntunes: Good call @ehmicky, we should handle file paths with spaces. This is probably a sign that we should do this another way to avoid bash sourceries related to word splitting also. I searched around a bit and from a page of `jq` faq - https://github.com/stedolan/jq/wiki/FAQ - we should probably rely on something like `mapfile -t array <<< $(jq -c '.[]' input.json)`

run-build-functions.sh
JGAntunes: @ehmicky we could use a global variable, this will work as long the `buildbot` runs these functions as part of the same shell execution I believe (which is the case now). It would however create a tighter coupling with the way we execute the `run-build.sh` script, meaning if we ever plan to install dependencies on a separate execution of which we cache things this would not work. That was my reasoning for using a file instead, but if you feel like it's too much happy to move to a different strategy.

run-build-functions.sh
JGAntunes: Handled via https://github.com/netlify/build-image/commit/ef73c8409763fd955bc04f8e1c1d1cb58e99a2b1

Running it locally:
```sh
NETLIFY_YARN_WORKSPACES feature flag set
yarn workspaces v1.22.4
{
  "gatsby-starter-blog-1": {
    "location": "packages/blog-1",
    "workspaceDependencies": [],
    "mismatchedWorkspaceDependencies": []
  },
  "gatsby-starter-blog-2": {
    "location": "packages/blog-2",
    "workspaceDependencies": [],
    "mismatchedWorkspaceDependencies": []
  },
  "blog-3": {
    "location": "packages/dir with spaces",
    "workspaceDependencies": [],
    "mismatchedWorkspaceDependencies": []
  }
}
Done in 0.12s.
Started restoring workspace packages/blog-1 node modules
Finished restoring workspace packages/blog-1 node modules
Started restoring workspace packages/blog-2 node modules
Finished restoring workspace packages/blog-2 node modules
Started restoring workspace packages/dir with spaces node modules
Finished restoring workspace packages/dir with spaces node modules
Started restoring workspace root node modules
Finished restoring workspace root node modules

(...)

Caching artifacts
Started saving workspace packages/blog-1 node modules
Finished saving workspace packages/blog-1 node modules
Started saving workspace packages/blog-2 node modules
Finished saving workspace packages/blog-2 node modules
Started saving workspace packages/dir with spaces node modules
Finished saving workspace packages/dir with spaces node modules
Started saving workspace root node modules
Finished saving workspace root node modules

(...)
```

ehmicky: :tada: 

Would `|` and `<` work instead of `<<< "$(...)"`? 

```bash
YARN_IGNORE_PATH=1 yarn --json workspaces info | jq -r '.data | fromjson | to_entries | .[].value.location' | mapfile -t package_locations
```

```bash
mapfile -t locations <"$NETLIFY_TMP_DIR/.workspace_locations"
```

Or (if we ended up using a variable):

```bash
mapfile -t locations <<<"$WORKSPACE_LOCATIONS"
```

run-build-functions.sh
ehmicky: Oh, I see what you mean.
I am wondering whether we will be in a situation where those different functions would end up in different processes.
Providing we are unsure about that, should we defer using IPC (through a file or other mean), and use a global variable instead?

run-build-functions.sh
JGAntunes: Piping won't work here unfortunately as it would create subshells to execute each part of it, meaning any changes to the `package_locations` variable will be local to that subshell - https://stackoverflow.com/a/18116407

run-build-functions.sh
JGAntunes: Yep 👍 I guess we can cross that bridge when it's time to. Addressed via https://github.com/netlify/build-image/commit/70c8a36e04ac440911ec776386d80536063768c6 @ehmicky

run-build-functions.sh
JGAntunes: No longer applicable, see - https://github.com/netlify/build-image/pull/526#discussion_r581135288

run-build-functions.sh
ehmicky: Oh, that's correct! :+1:

run-build-functions.sh
```diff
@@ -779,6 +817,56 @@ restore_cwd_cache() {
...
+  local cache_dir="$NETLIFY_CACHE_DIR/js-workspaces"
+
+  # Retrieve each workspace node_modules
+  for location in "${locations[@]}"; do
+    move_cache "$cache_dir/$location/node_modules" "$NETLIFY_REPO_DIR/$location/node_modules" "restoring workspace $location node modules"
+  done
+  # Retrieve hoisted node_modules
+  move_cache "$cache_dir/node_modules" "$NETLIFY_REPO_DIR/node_modules" "restoring workspace root node modules"
+  # Keep a record of the workspaces in the project in order to cache them later
+  NETLIFY_JS_WORKSPACE_LOCATIONS=$(printf '%s\n' "${locations[@]}")
```
ehmicky: Would it work to keep `locations` as a global variable directly instead of using `NETLIFY_JS_WORKSPACE_LOCATIONS` as an intermediate variable?

Like this, `cache_js_workspaces` could use it directly, without having to use `mapfile`. And `restore_js_workspaces_cache` would not need to use `printf`.

run-build-functions.sh
```diff
@@ -779,6 +817,56 @@ restore_cwd_cache() {
...
 
+#
+# Restores node_modules dirs cached for js workspaces
+# See https://github.com/netlify/pod-workflow/issues/139/ for more context
+#
+# Expects:
+# $@ each argument should be a package location relative to the repo's root
+restore_js_workspaces_cache() {
+  local locations=("$@")
+  local cache_dir="$NETLIFY_CACHE_DIR/js-workspaces"
```
ehmicky: Should this be moved to the top-level scope? To avoid it being declared twice.

run-build-functions.sh
JGAntunes: Good point @ehmicky! Addressed via https://github.com/netlify/build-image/commit/32f7b0df2c181f9dfeb4f654b1115ead254e46cf.

run-build-functions.sh
JGAntunes: Yeah I see what you mean, I ended up going with that pattern as it was something already being done for the other cache directories. But makes sense IMO. Addressed via https://github.com/netlify/build-image/commit/14b1ea8cb00c87c1b649b344b5ebd6f94683aedc.

run-build-functions.sh
```diff
@@ -102,6 +105,34 @@ run_yarn() {
...
+      # {
+      #   "my-package-1": {
+      #     "location": "packages/blog-1",
+      #     "workspaceDependencies": [],
+      #     "mismatchedWorkspaceDependencies": []
+      #   },
+      #   (...)
+      # }
+      # We need to cache all the node_module dirs, or we'll always be installing them on each run
+      mapfile -t package_locations <<< "$(YARN_IGNORE_PATH=1 yarn --json workspaces info | jq -r '.data | fromjson | to_entries | .[].value.location')"
```
ehmicky: `YARN_IGNORE_PATH=1 yarn workspaces info` is called in the test above. 
Would it make sense to make the test use `--json` and keep the output, to avoid calling `yarn workspaces info` twice? The test would then check exit code using `$?`. As a tiny performance optimization.

ehmicky: Would it work to directly set `NETLIFY_JS_WORKSPACE_LOCATIONS` instead of `package_locations`? This might require declaring `NETLIFY_JS_WORKSPACE_LOCATIONS` as a file-level array variable. This would remove the need for the intermediary variable `package_locations`.

JGAntunes: Right now the cache logic and the respective usage of `NETLIFY_JS_WORKSPACE_LOCATIONS` is kind of isolated from the `run_yarn` function, I would say there's some use in keeping it that way? 🤔 That way we're keeping all the convoluted logic required to cache `workspaces` in the respective `restore_js_workspaces_cache`, `cache_js_workspaces` and `cache_node_modules`. The only thing the `run_yarn` is doing right is to pass the `locations` to the underlying functions ensuring it's decoupled from that. My concern is that if we end up spreading the `NETLIFY_JS_WORKSPACE_LOCATIONS` variable around we might end up with a tighter spaghetti bowl 😅

JGAntunes: Addressed via https://github.com/netlify/build-image/pull/526/commits/073a3db99dd3523a392c1966071442f0f0bf1cdc 👍

ehmicky: Makes sense :+1:

JGAntunes: Now, actually fixed via https://github.com/netlify/build-image/pull/526/commits/8ab88aec42aabe2ec9ecb9681fc0c0a699189d87 😅

Btw, I'm keeping the exit code in a local variable as a way to avoid hitting - https://github.com/koalaman/shellcheck/wiki/SC2181 - let me know what you think @ehmicky

ehmicky: Sounds good!

run-build-functions.sh
```diff
@@ -102,6 +105,38 @@ run_yarn() {
   fi
 
 
+  if [ "$NETLIFY_YARN_WORKSPACES" = "true" ]
+  then
+    echo "NETLIFY_YARN_WORKSPACES feature flag set"
+    local workspace_output
+    local workspace_exit_code
+    # YARN_IGNORE_PATH will ignore the presence of a local yarn executable (i.e. yarn 2) and default
+    # to using the global one (which, for now, is alwasy yarn 1.x). See https://yarnpkg.com/configuration/yarnrc#ignorePath
```
```suggestion
    # to using the global one (which, for now, is always yarn 1.x). See https://yarnpkg.com/configuration/yarnrc#ignorePath
```

run-build-functions.sh
```diff
@@ -102,6 +105,38 @@ run_yarn() {
...
+    workspace_output="$(YARN_IGNORE_PATH=1 yarn workspaces --json info )"
+    workspace_exit_code=$?
```
ehmicky, line 116: [sand!]

```suggestion
    local workspace_exit_code=$?
```

(And removing the declaration above)
(Same for `workspace_output`)

ehmicky (dismissed):
:tada:

run-build-functions.sh
JGAntunes: @ehmicky happy to do so for the `workspace_exit_code` however, for the previous one, declaring a `local` variable will actually override the exit code from the command we need to use afterwards - https://github.com/koalaman/shellcheck/wiki/SC2155

ehmicky: TIL.
Good catch :+1:

ehmicky (approved):

run-build-functions.sh
JGAntunes: I owe it all to shellcheck - https://github.com/koalaman/shellcheck/ - I would be doomed writing bash/sh without it 😂

PR #529, opened 2021-03-08
fix: changelog and pacakge.json versioning

JGAntunes:
Without being sure this is 100% correct, it brings our changelog as well as our `package.json` up to date with what we have in our `xenial` branch.


ehmicky (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@netlify/build-image",
-  "version": "3.4.4",
+  "version": "3.6.1",
```
ehmicky, line 3: What are your thoughts on removing `package-lock.json` from `.gitignore` (separately)?

JGAntunes: Yeah, I guess we could do it 👍

PR #530, opened 2021-03-08
chore: add release please

JGAntunes:
Brings this project in line with what we're using in our other projects for release purposes 👍 makes releases easier and we avoid issues such as #529

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,14 @@
+ name: release-please
+ on:
+   push:
+     branches:
+       - master
+ jobs:
+   release-please:
+     runs-on: ubuntu-latest
+     steps:
+       - uses: GoogleCloudPlatform/release-please-action@v2
+         with:
+           token: ${{ secrets.GITHUB_TOKEN }}
+           release-type: node
+           package-name: "@netlify/build-image"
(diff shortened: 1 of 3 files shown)


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,14 @@
...
+   push:
+     branches:
+       - master
+ jobs:
+   release-please:
+     runs-on: ubuntu-latest
+     steps:
+       - uses: GoogleCloudPlatform/release-please-action@v2
+         with:
+           token: ${{ secrets.GITHUB_TOKEN }}
```
ehmicky: Note: we are using `NODE_PKG_RELEASE_TOKEN` in other repositories. 

https://github.com/netlify/cli/blob/073217f558411e92f4dc29793308c616649ad41f/.github/workflows/release-please.yml#L12

JGAntunes: Probably makes sense to do so here, else our CI workflow won't be triggered by the `release-please` bot 😕  but I'll need a repo admin to add that secret 👀

JGAntunes: Feels like https://github.com/netlify/engineering/issues/259 would be the ideal solution though.

ehmicky (approved):

PR #532, opened 2021-03-09
fix(release-please): set the correct branch name

JGAntunes:
Branch naming gets me every time 😄 fixes #530

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -2,7 +2,7 @@
  on:
    push:
      branches:
-       - master
+       - xenial
  jobs:
    release-please:
      runs-on: ubuntu-latest


.github/workflows/release-please.yml
```diff
@@ -2,7 +2,7 @@
  on:
    push:
      branches:
-       - master
+       - xenial
```
ehmicky, line 5: Should there be releases for the other Ubuntu versions?
Maybe not, but I was curious about what you thought.

ehmicky (approved):

.github/workflows/release-please.yml
JGAntunes: Good question 🤔 I guess we're not actively releasing to it, and the only `trusty` related tag we seem to have in docker hub is well... trusty, so it seems to me we're only backporting changes whenever we have a need to do so. I'll go ahead with the `xenial` branch only, we can revisit this later if we need to 👍

vbrown608: +1, Trusty is released only for critical security fixes :+1:

PR #540, opened 2021-03-25
syntax: bash launcher script hardening

hraban:
Apply various hardening idioms to the default Bash launcher script:

- quote all variables
- quote all path names
- explicit error message on wrong usage
- use exec for the "wrapped" command
- use common hardening flags (this one is not technically useful in this script as it is, but it's a good idea anyway, in case it is ever modified with a statement that could benefit from it)
- use /usr/bin/env


kitop:

PR #543, opened 2021-04-08
feat: upgrade to Ubuntu 20.04 Focal Fossa

ingride:
Upgrading to the current [LTS release - 20.04 Focal Fossa](https://releases.ubuntu.com/).

A few updates along with this version upgrade

**Deprecations**:
- PHP versions 5.2 to 7.2 that have reached EOL
- Python 3.5 and 3.7 are no longer supported by Ubuntu 20.04
- npm `sm` package due to lack of development and new releases over past 8years
- WIP `libcurl3-openssl-dev` due to a conflict with `libcurl4-gnutls-dev`. since `libcurl4` is an openssl flavour we need to test if this is enough
- gnupg-curl does not have Focal support and has been removed 

**Explicit Upgrades**:
- libcurl3 upgrades to libcurl4
- hugo since we were quite a few versions behind
- ruby 
- elm
- yarn
-  libwebp5 to  libwebp6
-  emacs-nox 
- libpng-dev 
- pandoc and wkhtmltopdf to latest versions

**Changes in the way we install packages:**
- git-lfs is now installed using `apt-get` 
- python is defaulted to 3.8 on Focal and doesn't need to be explicitly installed 
- Libvips is now installed as an apt package
- PHP only has one supported version atm - 8 is not ready yet for Focal - so no need to default to one anymore 

Tracking issue: https://github.com/netlify/pod-workflow/issues/148


Dockerfile
```diff
@@ -217,28 +199,6 @@ RUN wget https://github.com/jgm/pandoc/releases/download/$PANDOC_VERSION/pandoc-
     rm pandoc-$PANDOC_VERSION-1-amd64.deb && \
     pandoc -v
 
-################################################################################
-#
-# Libvips
```
ingride, line 222: we have this as an Ubuntu package now, no need to reinvent the wheel.

verythorough:
Awesome!

Notes on in-repo docs that will need to be updated in this branch:
- included-software.md ([Xenial version](https://github.com/netlify/build-image/blob/xenial/included_software.md))
- README: add Focal to the list of images available, as an opt-in beta for now. We'll want to update this on all three release branches. When we have a deprecation schedule set, we can add that as well.

Dockerfile
```diff
@@ -120,41 +113,28 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
...
-        php5.6-zip \
-        php5.6-intl \
-        php7.2 \
-        php7.2-xml \
-        php7.2-mbstring \
-        php7.2-gd \
-        php7.2-sqlite3 \
-        php7.2-curl \
-        php7.2-zip \
-        php7.2-intl \
```
vbrown608, line 157: > PHP only has one supported version atm - 8 is not ready yet for Focal - so no need to default to one anymore

Nice to see all these deprecated versions getting attention!

Looks like the install script, which runs as the `buildbot` user, doesn't auto-install any PHP versions, just selects one of the preinstalled versions based on https://github.com/netlify/build-image/blob/xenial/run-build-functions.sh#L374-L388

Since this PR removes all preinstalled versions, what do you think about researching some installation method that works without root so that users can install the version they need at runtime? I wonder if homebrew would be a good alternative.

verythorough:
@ingride Do you think we could incorporate the proposed changes in https://github.com/netlify/build-image/pull/535 into the new image? Bascially bumping pandoc and wkhtmltopdf to current versions.

Dockerfile
ingride: We still have PHP 7.4 which is [the only version supported by Ubuntu 20.04 at the moment](https://packages.ubuntu.com/search?suite=focal&arch=any&searchon=names&keywords=php). Versions lower than 7.4 haven't been ported to Focal because they have reached EOL or [end of support(7.3)](https://www.php.net/supported-versions.php) . It actually seems [7.3 never had a package](https://www.php.net/supported-versions.php). I suspect 8.0 will be released soon and we can add it then.

vbrown608: Ah, I missed that we still have that one :+1: 

I don't know enough about PHP to know if we have a lot of users on old versions and what the upgrade path would be like for them. I do think providing some auto-install at runtime would be cool but we'll know more about whether PHP versions are a pain point once we start beta testing the image, so this seems fine for now.

ingride: ➕1 to iterate as we get feedback for this image and if needed we can consider manually installing or as you suggested doing something at runtime.

ingride:
@verythorough good point! I was hoping we could get these from ubuntu images - especially pandoc, but it seems they're far behind still :( . I updated to the latest using the old way.

ingride:
tested the image with different sites for covering a range of dependency upgrades, and it's looking good https://github.com/netlify/buildbot/pull/1284#issuecomment-825068543 

i think we can review this and get it merged and keep iterating as per the release plan https://github.com/netlify/pod-workflow/issues/199

README.md
```diff
@@ -12,6 +12,7 @@ The following images are currently available:
 
 - `trusty` - Legacy build image for older sites; running Ubuntu 14.04 and [this software](https://github.com/netlify/build-image/blob/trusty/included_software.md)
 - `xenial` - Default build image for all new sites; running Ubuntu 16.04 and [this software](https://github.com/netlify/build-image/blob/xenial/included_software.md)
+- `focal` - Opt-in beta build image; running Ubuntu 20.04 and [this software](https://github.com/netlify/build-image/blob/xenial/included_software.md)
```
verythorough: Until we actually have this available for selection in the UI, I wonder if we change this line to something like this?
```suggestion
- `focal` - In development and available for local testing; running Ubuntu 20.04 and [this software](https://github.com/netlify/build-image/blob/xenial/included_software.md)
```
I say "local testing" to refer to running the steps in the readme with the `focal` branch named in the `docker pull` command.

When we have it available in the UI, we can change to the "opt-in beta" language.

included_software.md
```diff
@@ -1,6 +1,6 @@
 ## Included Software: Xenial
```
verythorough:
```suggestion
## Included Software: Focal
```
😄

ingride:
thanks @verythorough!! ❤️

Jenkinsfile
```diff
@@ -4,7 +4,7 @@ pipeline {
   stages {
     stage("Test Build") {
       when {
-        not { anyOf { branch 'staging' ; branch 'xenial' ; branch 'trusty  ' ; buildingTag() } }
```
vbrown608, line 7: Should we include `focal` in this list?

vbrown608 (commented):
This looks good to me! Just had one dust-ish question.

It's hard to evaluate individual dependencies without deep knowledge of all of the technologies involved, so thank you for sharing the plan for internal testing and incremental rollout. I think this will be a good base to iterate from.

Benaiah (commented):
Looks good! Once @vbrown608's [comment](https://github.com/netlify/build-image/pull/543/files#r618745278) is addressed this seems ready to go to me.

Jenkinsfile
ingride: yes, i intended to but seems it got missed. thanks for catching this.

vbrown608 (approved):
:framed_picture: :tada:

README.md
```diff
@@ -12,6 +12,7 @@ The following images are currently available:
 
 - `trusty` - Legacy build image for older sites; running Ubuntu 14.04 and [this software](https://github.com/netlify/build-image/blob/trusty/included_software.md)
 - `xenial` - Default build image for all new sites; running Ubuntu 16.04 and [this software](https://github.com/netlify/build-image/blob/xenial/included_software.md)
+- `focal` - In development and available for local testing; running Ubuntu 20.04 and [this software](https://github.com/netlify/build-image/blob/xenial/included_software.md)
```
ab, line 15: `xenial` should presumably be `focal` in the link here, thanks!

PR #545, opened 2021-04-23
fix: remove duplicated go import path code

Benaiah:
Removes duplicated code and no longer deletes siblings of the import path we're adding.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -891,6 +891,7 @@ set_go_import_path() {
     local dirPath="$(dirname $importPath)"
 
     rm -rf $importPath
+    mkdir -p $dirPath
     ln -s $PWD $importPath
 
     cd $importPath
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -898,8 +898,7 @@ set_go_import_path() {
     local importPath="$GOPATH/src/$GO_IMPORT_PATH"
     local dirPath="$(dirname $importPath)"
 
-    rm -rf $dirPath
-    mkdir -p $dirPath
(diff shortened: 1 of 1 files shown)


Benaiah:
It looks like `fix: ` is supposed to be a valid Conventional Commits prefix so I'm not sure why the CI step linting that is failing.

ehmicky:
@Benaiah `release-please` is using the GitHub PR title.

JGAntunes (approved):

ingride (approved):
can you please also port this to focal 

🚀

ingride (commented):
actually, seems jenkins pr-head is unhappy.

run-build-functions.sh
```diff
@@ -898,7 +890,7 @@ set_go_import_path() {
     local importPath="$GOPATH/src/$GO_IMPORT_PATH"
     local dirPath="$(dirname $importPath)"
 
-    rm -rf $dirPath
+    rm -rf $importPath
```
Benaiah, line 893: Looking at this again, I'm actually not sure this is removing the correct directory anymore. This was removing `$PWD/dirname`, but it's now removing `$GOPATH/src/$GO_IMPORT_PATH` instead. I think this should remain as `rm -rf $dirpath` like it was before.

Benaiah: Nevermind - I changed that on purpose and I'm now remembering why. Before, it was removing siblings of the `$importPath` and now it's *only* removing the `$importPath` itself.

Benaiah:
Created #545 as a duplicate of this PR but against Focal.

PR #547, opened 2021-04-26
fix: Update homebrew install

mraerino:
The Dockerfile was using an old fork by David (former CTO) which installed a homebrew version that no longer worked.

I switched the Dockerfile to use the official script and swapped some things around to make it work.

i ran this to verify locally:
```
% docker build . -t netlify/build-image
% docker run -it --rm netlify/build-image bash
$ brew install capnp
```


ingride (approved):
🏡

fasterthanlime (approved):

PR #548, opened 2021-04-26
chore: update focal in ci and release-please scripts

ingride:
This PR updates the release-please workflow along with ci and tests. 

Part of releasing focal: https://github.com/netlify/pod-workflow/issues/199

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,4 +13,4 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           currentTag: netlify/build:${{ github.sha }}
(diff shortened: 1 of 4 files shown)


ehmicky (approved):

vbrown608:
For a follow-up, what do you think about updating the README to explain how to test with both the default (focal) image and the xenial image?

https://github.com/netlify/build-image/blob/xenial/README.md#step-1-pull-the-build-image

Since `release-please` won't run on merges to the `xenial` branch, would we tag 3.x releases manually on that branch going forward? I'm wondering what would would happen if we run it on merges to both `xenial` and `focal`.

PR #551, opened 2021-04-28
chore: default php version to 5.6 in xenial

ingride:
Currently, in buildbot, we don't actually check if there is a PHP version in the user's environment - we default to 5.6 for xenial. 
With us starting to [build both focal and xenial from the same main branch](https://github.com/netlify/buildbot/pull/1303#discussion_r622363195), we can move this default to the actual image.

This PR unblocks: https://github.com/netlify/buildbot/pull/1303


vbrown608 (approved):
Looks good! :elephant: :elephant: (I googled the PHP mascot :grin: )

PR #552, opened 2021-04-28
chore: default php version to 7.4 in Focal

ingride:
Currently, in buildbot, we don't actually check if there is a PHP version in the user's environment - we default to 5.6 for xenial, but this is a version no longer supported in Focal (https://github.com/netlify/pod-workflow/issues/187).
With us starting to build both focal and xenial from the same main branch, we can move this default to the actual image.

This PR unblocks: netlify/buildbot#1303


vbrown608 (approved):
:elephant: :seven: :four:

Issue #485, opened 2020-10-27
Cache dependencies for clojure projects using Tools.deps.alpha in addition to Leiningen and boot


armincerf:
Currently Netlify will only cache dependencies in Clojure projects if there is a valid `project.clj` or `build.boot` file, which are used by the Leiningen and Boot build tools.

However Clojure now has an official API for dependency management and uses a `deps.edn` file rather than the `project.clj` or `build.boot` files used in older build tools.

It makes sense to add support for `deps.edn` files as this is the official way to manage dependencies in Clojure now.

Relevant PR https://github.com/netlify/build-image/pull/117/files

kennyjwilli:
Are there any known workarounds for the time being?

armincerf:
The only one I can think off is maintaining a project.clj as well, but that's a bit of a pain

armincerf:
Actually just looked at the code and seems like it doesn't actually matter what's in the project.clj file, just that it exists. I'll test with an empty project.clj to make sure tomorrow but if that works then it will be easy to check for project.clj or deps.edn as they will both be cached in the same way.

I will make a PR when I get around to it if that's the case.

kennyjwilli:
I think that will work well for most cases. Our build currently eats up ~2.5 mins of downloading deps on every build haha. 

The one exception is that deps.edn allows for Git deps which are stored in `~/.gitlibs` (by default. [See docs](https://clojure.org/reference/deps_and_cli#_git)).

minikomi:
A completely empty (`touch project.clj`) project.clj causes lein to try and run it, which results in an error.

mainej:
I believe this would be fixed by #553.

kennyjwilli:
Hi @mainej. Thanks for working on that PR. The Clojure CLI supports [Git Deps](https://clojure.org/reference/deps_and_cli#_git). I'd suggest adding the `~/.gitlibs` directory to the cached set of dirs ([see docs](https://clojure.org/reference/deps_and_cli#_configuration_and_debugging)).

mainej:
@kennyjwilli Good suggestion. 👍 I've added caching of `~/.gitlibs` to the PR and verified that it works by using the [instructions for testing builds with the cache](https://github.com/netlify/build-image#testing-locally-with-cache).

barrelltech:
~What is the status of this? I would like to use netlify but am building a project with the clojure cli~

[...]

PR #553, opened 2021-04-29
feat: Cache Clojure dependencies

mainej:
This PR adds caching for repos built with the [Clojure CLI tools][clojure-cli]. It extends 52f8eb9, which added initial support for building with the Clojure CLI tools, but did not address caching. It is similar to f3a4d49 which introduced caching for Leiningen and Boot, two other ways of building Clojure code.

This PR contains three commits. I'd like to see them all put into use, but the second one is certainly optional.

The first commit will significantly improve build times and decrease network bandwidth for Clojure repos. It does this by restoring the `.m2` directory before building, as other `JVM` based tools do.

The second commit is less important. It has two parts. 
* First, it takes advantage of a cache that the Clojure CLI tools create, a cache which helps them avoid re-calculating a JVM classpath. In practice, I haven't seen any projects where that calculation is so slow that this cache will have much impact. 
* Second, it downloads dependencies to the `.m2` directory before starting the build. Technically this is unnecessary. As opposed to a tool like, for example, `bundler`, which expects to resolve and download dependencies before the "real" code is run, the Clojure CLI tools combine dependency management and code execution into one step. Therefore existing builds that use the Clojure CLI tools already populate `.m2`. So, this change merely separates the steps, isolating errors that occur during download.

The third commit caches the `~/.gitlibs` directory, which the Clojure CLI tools use to cache dependencies [procured from git repos][git-procurement] (as opposed to from package managers). This will improve build times and decrease network bandwidth for Clojure repos which use this type of dependency procurement.

[clojure-cli]: https://clojure.org/reference/deps_and_cli
[git-procurement]: https://clojure.org/reference/deps_and_cli#_git

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -350,6 +350,8 @@ RUN mkdir /opt/boot-clj && cd /opt/boot-clj && \
     chmod +x boot && \
     ln -s /opt/boot-clj/boot /usr/local/bin/boot
 
+# TODO: When upgrading to >= 1.10.1.672, also update approach for populating .m2 with clojure deps:
+# In run-build-functions.sh replace `clojure -Spath -Sforce >/dev/null` with more modern `clojure -P`
 RUN curl -sL https://download.clojure.org/install/linux-install-1.10.1.492.sh | bash
 
 USER buildbot
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -581,6 +581,12 @@ install_dependencies() {
     fi
   fi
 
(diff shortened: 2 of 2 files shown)


mainej:
I believe this would fix #485.

minikomi:
Would love to see this merged!

kitop:

mainej:
I've reopened this as #831, against the focal branch. I also took the opportunity to simplify the PR. /cc @minikomi

PR #555, opened 2021-04-29
feat(node): change the default node version

JGAntunes:
This is part of the work discussed in - https://github.com/netlify/pod-workflow/issues/180. We're also changing the current behaviour to only pin the major and minor version, so that new build-image builds always get the latest patches.

Setting this as draft for the time being while we settle some outstanding questions in https://github.com/netlify/pod-workflow/issues/180 and https://github.com/netlify/bitballoon/pull/9561

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -292,7 +292,7 @@ RUN git clone https://github.com/creationix/nvm.git ~/.nvm && \
 ENV ELM_VERSION=0.19.0-bugfix6
 ENV YARN_VERSION=1.22.4
 
-ENV NETLIFY_NODE_VERSION="12.18.0"
+ENV NETLIFY_NODE_VERSION="14.16"
 
 RUN /bin/bash -c ". ~/.nvm/nvm.sh && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -292,7 +292,7 @@ RUN git clone https://github.com/creationix/nvm.git ~/.nvm && \
 ENV ELM_VERSION=0.19.0-bugfix6
 ENV YARN_VERSION=1.22.4
 
-ENV NETLIFY_NODE_VERSION="12.18.0"
+ENV NETLIFY_NODE_VERSION="14.16"
```
ehmicky, line 295: Should we only pin the major instead?
Otherwise, we might miss some security patches, if the version is stuck to a previous minor version.

For example, if `14.17.0` was released with some non-breaking features, then `14.17.1` was released with some security fixes, we would miss those by pinning to `14.16`.

ehmicky:
Great idea to combine both of those in one PR! :heart:

JGAntunes:
Closing this in favour of - https://github.com/netlify/build-image/pull/588

PR #573, opened 2021-05-22
Fix link typo

johtso:


erezrokah (approved):
Thanks @johtso! Great catch :)

Issue #577, opened 2021-06-08
Use current supported language versions by default in Focal branch


verythorough:
I noticed that some of the default language versions in Focal's [included software list](https://github.com/netlify/build-image/blob/focal/included_software.md) are EOL. I think it's great to provide ways for users to run/install older versions to help ease migration from Trusty and Xenial, but it would be great to get new sites started with current supported versions.

Here are a few that I've noted:
- [Ruby 2.6 is still supported](https://endoflife.date/ruby), but our default version is 2.6.2, while the latest patch is 2.6.7. Is it possible/advisable to pin Ruby to minor versions instead of a specific patch?
- [Node 10 reached full EOL in April](https://endoflife.date/nodejs). Current supported versions are 14 and 16. Active support for 14 ends in October, but 16 was just released in April. I'm thinking 14 is probably safer for now, but folks in the dev tooling guild would have more expertise.
- [Python 2.7 reached EOL in Jan 2020](https://endoflife.date/python). The other version we have on Focal is 3.8, which will be good all the way until Oct 2024! It seems like a good version to default to, and we'll still have 2.7 as an option for migration support.
- [Go 1.12 hit EOL in Feb 2019](https://endoflife.date/go). Current supported versions are 1.15 and 1.16. Gophers on the team might have thoughts about which version makes most sense here. To be honest, I'm not sure how much people actually use the Go runtime in builds, anyway, since they often use compiled binaries (as with Hugo, for example).
- [Elixir 1.9 is not the current version](https://hexdocs.pm/elixir/compatibility-and-deprecations.html), but it does still get security patches, so maybe we just keep this one for the sake of stability, and consider upgrading if we get an issue for it.

In some cases, it may be that the included software list is just out of sync with the actual behavior, in which case we can just update the doc. In others, I think we'll want to update the build script, too. Let me know if I'm missing something!

ingride:
Imma reassign this to myself when i actually start working on it, just in case someone else might wanna do it

Benaiah:
@ingride @verythorough how do you feel about splitting this into separate issues, one for each of the dependencies?

ingride:
I don't mind, you might wanna do them all at once tho because the build image takes so long to build, test, and release that multiple PRs are actually a disadvantage here.

ingride:
> Ruby 2.6 is still supported, but our default version is 2.6.2, while the latest patch is 2.6.7. Is it possible/advisable to pin Ruby to minor versions instead of a specific patch?

[...]

PR #576, opened 2021-06-03
fix: php/iptables support for xenial and ruby and yarn upgrades 

ingride:
We currently pin versions on the repo level (https://github.com/netlify/bitballoon/issues/9481) so getting Focal out of beta would mean we'd need to make some changes to improve that behavior. 
Since that is a bigger lift, what we can do to get the ball rolling is keep some of the versions in sync: ruby and yarn

This PR: 
*  removes the WARN message when installing Ruby. Ruby is actually the only message we print a warning for, which is confusing for the user since we pre-install Ruby even if the current site is not using it. https://github.com/netlify/bitballoon/issues/9481 . We do give feedback as to which version has been pinned and will be installed, and removing the WARN will align the behavior with what we do for other languages. 
* fixes an issue with the PHP PPA deprecating all PHP versions for xenial. To mitigate this, we will now install these dependencies from local packages: tracking issue: https://github.com/netlify/buildbot/issues/1419 
* moves and installs iptables from source: details: https://github.com/netlify/buildbot/issues/1419#issuecomment-861389905

This should be followed up by a bitballoon PR to align defaults for ruby and yarn https://github.com/netlify/bitballoon/pull/9908. 

Tested with PHP site: https://app.netlify.com/sites/elegant-heyrovsky-57fbb5/deploys/60c8f1d24e42f000bb017ab1

PR #588, opened 2021-07-06
feat: upgrade default software versions in Focal

Benaiah:
Upgrades:

- Python 3: updates default from 3.8 (which is the default present in the system before installing packages) to 3.9, and uses 3.9 to install `pipenv` (instead of 2.7).
- Ruby: updates default from 2.6 to 2.7, and additionally installs 3.0.
- Go: updates default from 1.12 to 1.16.
- Node: updates default from 12.18 to 16 (suggested by @ehmicky).
- NVM: updates from 0.35 to 0.38.
- Swift: updates default from 5.2 to 5.4.
- Hugo: updates from 0.82 to 0.85.

Additionally removes an unused emacs PPA from the apt sources (we are installing `emacs-nox`, which is not a package from that PPA - it provides `emacs25`, `emacs26`, and `emacs27`).

In order to use these new defaults, we will also need to change the versions passed into `install_dependencies` in `run-build-functions.sh` by the build system.


included_software.md
```diff
@@ -9,19 +9,21 @@ For instructions on how to run this image locally to test your build, please see
...
+  * 2.6
+  * 3.0
   * Any version that `rvm` can install.
 * Node.js - `NODE_VERSION`, `.nvmrc`, `.node-version`
-  * 10 (default)
+  * 16.4 (default)
   * Any version that `nvm` can install.
 * Python - `PYTHON_VERSION`, `runtime.txt`, `Pipfile`
-  * 2.7 (default)
+  * 3.9 (default)
```
verythorough, line 20: You mentioned in the PR description that 3.8 is "the default present in the system before installing packages". Does that mean that 3.8 is available automatically without explicitly installing it? If so, I wonder if it still makes sense to stay with 3.8 for default, since it seems like that could reduce startup time? We still have [3 years](https://endoflife.date/python) before it hits EOL, which is almost as long as [Focal itself](https://endoflife.date/ubuntu).

included_software.md
```diff
@@ -9,19 +9,21 @@ For instructions on how to run this image locally to test your build, please see
...
 * Ruby - `RUBY_VERSION`, `.ruby-version`
-  * 2.6.2 (default)
-  * 2.7.2
+  * 2.7 (default)
+  * 2.6
+  * 3.0
   * Any version that `rvm` can install.
 * Node.js - `NODE_VERSION`, `.nvmrc`, `.node-version`
-  * 10 (default)
+  * 16.4 (default)
```
verythorough: Based on this [issue comment](https://github.com/netlify/build-image/issues/577#issuecomment-867988431), I think we're better off pinning Node to the major version (i.e. `16`) rather than to a minor.

verythorough (approved):
Thanks for this, @Benaiah! I added a couple of comments about possible changes to the versions.

verythorough (commented):
Oops, I accidentally ticked the "Approve" button instead of "Comment", and I have no idea how to change it. 🙈

bakerkretzmar:
I'd love to see PHP included here too and I'm happy to help out, should that be added to this PR or submitted as another?

I believe the default (and only) PHP version included in the Focal image is 7.4, which should probably be kept as an option, but the current version is 8.0.

ingride:
PHP 7.4 is already installed in the image as **bakerkretzmar** mentioned https://github.com/netlify/build-image/blob/focal/Dockerfile#L139,  tho [the active support is going to end in 4months](https://www.php.net/supported-versions.php). 

@Benaiah 8.0 doesn't look ready for focal yet from the [main Ubuntu package manager](https://packages.ubuntu.com/search?suite=focal&keywords=php8.0 ), but we could potentially look into using the[ PPA again](https://launchpad.net/%7Eondrej/+archive/ubuntu/php/+index?batch=75&memo=225&start=225) as we did for xenial before - https://github.com/netlify/build-image/pull/576/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L28. 

Alternatively, we can leave 8.0 out of the scope of this PR and follow up with a new one shortly. @bakerkretzmar, contributions are always welcomed!

Dockerfile
```diff
@@ -260,19 +262,24 @@ USER root
 #
 ################################################################################
 
-ENV PIPENV_RUNTIME 2.7
+ENV PIPENV_RUNTIME 3.9
+
+USER root
+
+RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 100 && \
```
ingride: i don't think we need 3.8 here right?

Dockerfile
```diff
@@ -145,8 +144,8 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         php7.4-zip \
         php7.4-intl \
         pngcrush \
-        python-setuptools \
         python3.8-dev \
```
ingride: you might wanna remove 3.8, no need to have both that and 3.9?

included_software.md
Benaiah: I've dug into this a bit and, while there doesn't appear to be any version of Python that's distributed in Ubuntu Focal by default, 3.8 is being installed already as a dependency of `software-properties-common`, which depends on `python3`, which in turn refers to Python 3.8. So, while I was wrong in thinking that it was present in the system *before* installing packages, it is going to be present in the system after installing that package, which we do at the top of the `Dockerfile` (see https://github.com/netlify/build-image/blob/focal/Dockerfile#L19).

Dockerfile
Benaiah: I can remove the explicit installation of `python3.8` here, and I'll do that now, but `python3.8` will still be present as it's a dependency of `software-properties-common` (which depends on `python3`, which installs `python3.8` and makes it the default provider of the `python3` binary).

Dockerfile
Benaiah: I added it as an alternative because it was also installed, but I don't think we really need this to be the case if we decide we want to use 3.9 - update-alternatives requires root to change the version used, so the only way someone could use python3.8 after these lines are run is by calling `python3.8` directly. I think it makes sense then to follow your suggestion and remove the line adding 3.8 as an alternative here.

Dockerfile
ingride: Good point on 3.8. we might wanna keep setuptools - some customers might be using it / depending on it?

Dockerfile
```diff
@@ -145,8 +144,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         php7.4-zip \
         php7.4-intl \
         pngcrush \
-        python-setuptools \
```
ingride, line 148: I responded to an outdated comment, but I think we should keep setuptools, might be useful for some customers and it's pretty popular.  any reason why we're removing it?

included_software.md
```diff
@@ -9,19 +9,21 @@ For instructions on how to run this image locally to test your build, please see
 The specific patch versions included will depend on when the image was last built (except Ruby). It is highly suggested you depend only on minor versions, so that we can ensure the language has the latest updates (especially if security related).
 
 * Ruby - `RUBY_VERSION`, `.ruby-version`
-  * 2.6.2 (default)
-  * 2.7.2
+  * 2.7 (default)
+  * 2.6
```
ingride: can we be more explicit about which 2.6 and 2.7 - for ruby it kinda matters.

ingride (commented):
smol things and a question. looks great otherwise 💚

Dockerfile
Benaiah: That sounds good, I'll leave setuptools in this PR.

Dockerfile
Benaiah: I didn't have a particular reason for removing it so I'll leave it in.

included_software.md
Benaiah: Sure, I'll specify which versions in the document.

Benaiah:
I've specified the full version of Ruby in `included_software.md` and added `python3-setuptools` to the image. The CI failure is due to an invalid signature related to Erlang - I'm not sure if this is something that will go away on its own or require a fix from our end, so I'm going to rerun CI.

Dockerfile
ingride: kewl, i think we should, as long as we still support python 2.7

ingride (approved):
so this looks good, but it's unfortunate the build image is almost 4.5% larger now. I suspect with adding PHP that will increase further. I think this is fine for the defaulting to focal, but what do you think of opening an issue so start reducing the image size further - maybe retire python 2.x and ruby 2.6.6 for the next iteration? 

Can you include some links with tests for ruby and python sites? assuming that went well, this is good to go. 

🏈

Benaiah:
I've decided to add `python-setuptools` back into this PR to make sure that Python 2 sites continue to work correctly after this change, since we're not removing Python 2 from the image entirely at this point.

I'll update shortly with links to ruby and python test sites as requested.

Benaiah:
@ingride I've created the issue you requested at https://github.com/netlify/build-image/issues/603.

Benaiah:
While preparing the Python test site, I ran into an issue where it looks like we do *not* install missing python versions, even if they are specified by the `PYTHON_VERSION` environment variable, the `requirements.txt`, or the `Pipfile`. This means that we need to continue to provide all the python versions that were previously available in order to allow sites to migrate to focal without source code changes. This shouldn't be as much of an issue when it comes to defaulting new sites to focal, but we may want to eventually change our dependency installation system to use `pipenv` to install new versions of python when we don't have the requested version available.

Benaiah:
I forgot to mention this in the last message, but the big issue with not installing missing python versions is that anyone who's already migrated their python site to focal and used a python version that's no longer available (such as 3.7) will have their builds start breaking once we release this change.

EDIT: here's an example of such a failed deploy: https://app.netlify.com/sites/deploy-canary-test-site-pelican/deploys/610870058de07f8860dda64d

Benaiah:
Here's an example of a working deploy using Python 3.9 and Pelican: https://app.netlify.com/sites/deploy-canary-test-site-pelican/deploys/610870b260c9ad0009f211ec

Benaiah:
I tried to run a build with 3.8 by setting the `Pipfile` to require 3.8, but it appears that it still used 3.9. I'm going to see if the `runtime.txt` method has the same issue. See https://app.netlify.com/sites/deploy-canary-test-site-pelican/deploys/610871622f788700089a75f9

Benaiah:
It looks like to allow python 3.8 sites to run we need to duplicate the logic here: https://github.com/netlify/build-image/blob/focal-upgrade-software-versions/Dockerfile#L279-L281 for python 3.8. It's already installed, but without the existing virtualenv it can't actually be used. This doesn't solve the issue of the `Pipfile` for a 3.8 version running in 3.9 instead - I'll look into that further.

Benaiah:
OK, I have a build using 3.8 working: https://app.netlify.com/sites/deploy-canary-test-site-pelican/deploys/610894f3422fca5f77ac729a

I'm going to work on a test case for python 2.7 and some for ruby.

Benaiah:
It looks like adding the 3.8 virtualenv fixed the issue where specifying 3.8 in the Pipfile didn't use 3.8 - see https://app.netlify.com/sites/deploy-canary-test-site-pelican/deploys/61089a2fe0c327000866f306

Benaiah:
I'm having trouble finding an SSG to test with that still supports Python 2.7 - is anyone aware of one that does? I'll keep looking in the meantime.

Benaiah:
Here's a working test site using Jekyll on (preinstalled) Ruby 2.7: https://app.netlify.com/sites/deploy-canary-test-site-jekyll/deploys/6108b4bcf5c20283d90c560e

ingride:
@Benaiah this seems like it supports python2 https://github.com/sunainapai/makesite

Benaiah:
Here's a working build using python 2: https://app.netlify.com/sites/deploy-canary-test-site-makesite/deploys/61099cc0c94e0500e65370ed

Benaiah:
Here's a working build using the preinstalled node (which is currently `16.6.0` in the branch I'm using for testing): https://app.netlify.com/sites/deploy-canary-test-site-gatsby/deploys/6109ae2f05d10400071b1861

And here is a successful build using a different Node version (`16.4.2`): https://app.netlify.com/sites/deploy-canary-test-site-gatsby/deploys/6109ada735b06900079bb695

ingride:
yay, amazing! thanks for running the tests @Benaiah! let's ship this ⛵

PR #592, opened 2021-07-15
feat: Install and default to PHP 8.0 (Focal)

bakerkretzmar:
This PR is a follow-up to https://github.com/netlify/build-image/pull/588#issuecomment-877420397 – it installs PHP 8.0 in the Focal build image and sets it to be the default version.

PHP 8.0 is the current version of PHP and has been since November 2020, and as @ingride noted in #588, active support for PHP 7.4 will end at the end of November 2021. See https://www.php.net/supported-versions.php.

I'm not super familiar with Docker or with this build image, I think I updated `7.4` to `8.0` everywhere I needed to but I may have missed places. I was able to build the image locally, run it, open a shell inside the running container, and verify that PHP 8.0 is installed correctly and is the default version.

Let me know if there's anything else I can do here, and feel free to fold this into #588 directly if that's easier!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     locale-gen en_US.UTF-8 && \
     update-locale en_US.UTF-8 && \
     apt-key adv --fetch-keys https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc && \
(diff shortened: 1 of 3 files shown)


Benaiah (approved):

ingride:
@Benaiah we might wanna have this merged before we cut the release for your changes as well.

bakerkretzmar:
Thanks so much!

Issue #593, opened 2021-07-15
`.yarn/cache` is never persisted between runs


JGAntunes:
Within `build-image` we rely on a specifc dir `.yarn_cache` which we create, pass to the `yarn install` command via the `--cache-folder` flag and then move to our caching directory (see - https://github.com/netlify/build-image/blob/xenial/run-build-functions.sh#L149)

We do however use `$NETLIFY_BUILD_BASE/.yarn_cache` and cache `$HOME/.yarn_cache` (see - https://github.com/netlify/build-image/blob/xenial/run-build-functions.sh#L706), which for `build-image` specifically is the same dir (see - https://github.com/netlify/build-image/blob/xenial/run-build-functions.sh#L10) however that's not the case when executing within `buildbot`'s context, as `$NETLIFY_BUILD_BASE` is overriden (see - https://github.com/netlify/buildbot/blob/b7aeb2675f82192daa8545b22bebab9c3b179d0e/conf/conf.go#L35 and https://github.com/netlify/buildbot/blob/master/bot/buildbot.go#L113.

Logs with a practical example:
- https://app.netlify.com/sites/paste-test-antunes/deploys/60f03f5871bb854864eb34f5#L5974
- https://app.netlify.com/sites/paste-test-antunes/deploys/60f03f5871bb854864eb34f5#L6463
- https://app.netlify.com/sites/paste-test-antunes/deploys/60f03f5871bb854864eb34f5#L7434

JGAntunes:
Closed via:
- https://github.com/netlify/build-image/pull/598
- https://github.com/netlify/build-image/pull/596

And released to prod 👍

PR #595, opened 2021-07-16
fix(yarn): make yarn install use the correct .yarn_cache path

JGAntunes:
Addresses #593. `.yarn_cache` is a `$HOME` folder, that for `build-image` runs coincides with `$NETLIFY_BUILD_BASE` but that's not the case for production builds.


eduardoboucas (approved):
🎸

PR #602, opened 2021-08-02
feat: cache directory for Rust functions (Focal)

eduardoboucas:
Saves to and loads from cache a `.netlify/rust-functions-cache`, if one exists. This directory will be used by zip-it-and-ship-it as the compilation target for Rust functions, allowing the bundling of these functions to be sped up between builds.

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -50,8 +50,10 @@ mkdir -p $NETLIFY_CACHE_DIR/.bundle
 mkdir -p $NETLIFY_CACHE_DIR/bower_components
 mkdir -p $NETLIFY_CACHE_DIR/.venv
 mkdir -p $NETLIFY_CACHE_DIR/.build
+
+# If you remove this line, please ensure that $NETLIFY_CACHE_DIR/.netlify is
+# still created, as it is used by the Rust functions cache.
 mkdir -p $NETLIFY_CACHE_DIR/.netlify/plugins
-mkdir -p $NETLIFY_CACHE_DIR/.netlify/rust-functions-cache
 
 # HOME caches
 mkdir -p $NETLIFY_CACHE_DIR/.yarn_cache
@@ -682,7 +684,7 @@ install_dependencies() {
     restore_cwd_cache "target" "rust compile output"
     source $HOME/.cargo/env
   fi
-  restore_cwd_cache ".netlify/rust-functions-cache" "Rust functions cache"
+  restore_cwd_cache ".netlify/rust-functions-cache" "Rust functions"
 }
 
 #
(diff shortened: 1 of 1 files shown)


run-build-functions.sh
```diff
@@ -697,6 +698,7 @@ cache_artifacts() {
   cache_cwd_directory ".venv" "python virtualenv"
   cache_cwd_directory ".build" "swift build"
   cache_cwd_directory ".netlify/plugins" "build plugins"
+  cache_cwd_directory ".netlify/rust-functions-cache" "Rust functions"
```
ehmicky, line 704: I am not sure whether there is a good reason for it, but it seems like every cached directory is created with `mkdir -p` at the top of the file.

https://github.com/netlify/build-image/blob/653805ca4a64301556e56dc4b321ef8fc20cbb7c/run-build-functions.sh#L53

I am wondering if this should be the case here as well?

ehmicky (dismissed):

run-build-functions.sh
eduardoboucas: I wondered about that myself. I left it out because I wanted to make sure we only print log messages about a Rust functions cache if people are indeed using Rust functions, otherwise they might feel like their builds are taking longer due to a feature they're not even using.

I believe the current logic will work because:

- `cache_cwd_directory` uses `move_cache` (https://github.com/netlify/build-image/blob/xenial/run-build-functions.sh#L847)
- `move_cache` checks whether the directory exists — if it does, it prints a log message and copies the directory (https://github.com/netlify/build-image/blob/xenial/run-build-functions.sh#L765)

I think the only risk of not using `mkdir -p` is if we try to copy file files to `$PWD/.netlify/rust-functions-cache` and `$PWD/.netlify` doesn't exist. At the moment, we know that it's being created [as part of the build plugins directory](https://github.com/netlify/build-image/pull/602/files#diff-6ad81dc23c4c92a6e6193db3b05e306293aa75ee25d48ae952c6259c601e8b23R54-R56), but I've added a comment to make it clear that it's also required for the Rust functions cache.

Does that make sense, or am I missing some detail here?

ehmicky: Makes sense! :+1:

ehmicky (dismissed):

run-build-functions.sh
```diff
@@ -50,6 +50,9 @@ mkdir -p $NETLIFY_CACHE_DIR/.bundle
 mkdir -p $NETLIFY_CACHE_DIR/bower_components
 mkdir -p $NETLIFY_CACHE_DIR/.venv
 mkdir -p $NETLIFY_CACHE_DIR/.build
+
+# If you remove this line, please ensure that $NETLIFY_CACHE_DIR/.netlify is
+# still created, as it is used by the Rust functions cache.
 mkdir -p $NETLIFY_CACHE_DIR/.netlify/plugins
```
ehmicky, line 56: Should we do a separate `mkdir -p $NETLIFY_CACHE_DIR/.netlify`?
This is unnecessary due to the `-p` flag, but I am wondering whether it might help avoid the problem mentioned in this comment.

```js
mkdir -p $NETLIFY_CACHE_DIR/.netlify
mkdir -p $NETLIFY_CACHE_DIR/.netlify/plugins
```

eduardoboucas: Much better! Done in https://github.com/netlify/build-image/pull/602/commits/4ccc5c322e7c09e49717d53a17ac63019c96f34e.

ehmicky (approved):

Issue #603, opened 2021-08-02
Reduce build image size by retiring old software versions


Benaiah:
The PR updating software versions in focal (#588) increases the build image size by ~4.5%. We may want to consider dropping old software versions such as Python 2 or Ruby 2.6 to reduce the size of the build image. Ideally, these would still be installable at build time for customers who require them, just not included by default in the build image.

verythorough:
I can see the motivation to remove older software versions, but Ruby 2.6 and Python 2.7 are the defaults in Trusty and Xenial, which means that nearly all existing sites are pinned to those versions, including, of course, sites that don't even use Ruby or Python. We don't have a way to install Python versions on demand, so if we removed 2.7, we would basically guarantee that nearly every upgrade to Focal would result in a failed build.

With this in mind, I don't think we can remove these versions until we change how versions are pinned. I'd like to prioritize doing https://github.com/netlify/pod-workflow/issues/255 as soon as possible as an initial improvement, and then we can consider other solutions as discussed in https://github.com/netlify/bitballoon/issues/9481.

On the other hand, I don't think we should consider this a blocker for defaulting new sites to using Focal. Xenial has been EOL for over 4 months, and customers are complaining that our only non-EOL image is still marked as "beta". Every new site that's created on Xenial now is another site that will have to be migrated to Focal when we deprecate Xenial later this year. A 4.5% larger build image is an acceptable tradeoff in the face of those issues.

verythorough:
Hm, perhaps there's one other alternative to reduce the image size without breaking so many builds. We could remove Python 3.9 and Ruby 2.7. The next highest versions on the image (Python 3.8 and Ruby 2.6) are supported, so it still meets the goal of defaulting to supported versions, while reducing the number of installed versions.

If the changes in https://github.com/netlify/build-image/pull/588 and https://github.com/netlify/bitballoon/pull/10222 had been released, this would be a breaking change for sites on Focal, but it seems they haven't been released yet, so we have a little window to make this change if image size reduction is a priority.

ingride:
@verythorough I don't think we should remove ruby 2.7 because we don't have a recent 2.x alternative that is not too old  so we would pay the install penalty for those sites on first run (2.6.7 is for supporting some customers that still need this version and can't update to 2.7). 3.x is still new enough, we don't even use it.  if we wanna remove something, I vote for ruby 3.x and adding it at a later time, and zapping 2.6.x

cc @Benaiah 

[...]

PR #611, opened 2021-08-10
chore: remove ruby 3 and python 3.9

Benaiah:
This came out of the conversation in https://github.com/netlify/build-image/issues/603, where we discussed removing ruby 3 and python 3.9 in order to minimize the additional size of the build image from new software versions.


included_software.md
```diff
@@ -17,8 +17,7 @@ The specific patch versions included will depend on when the image was last buil
   * 16 (default)
   * Any version that `nvm` can install.
 * Python - `PYTHON_VERSION`, `runtime.txt`, `Pipfile`
-  * 3.9 (default)
```
ingride, line 20: ruby 3.0 should also be ⚡

ingride (commented):
looks good otherwise.

included_software.md
Benaiah: great catch, thank you!

Benaiah:
I've fixed the doc issue you mentioned and added a commit updating the default versions used in `run-build.sh` as per https://github.com/netlify/build-image/pull/605.

ingride (approved):

ingride (approved):
🐍

straight-shoota:
I don't really follow why remvoing the most recent Python release would be considered a good idea. If you want to save space, I'd recommend throwing out *old* releases.

PR #616, opened 2021-09-01
feat(install_dependencies): improve logging of Ruby, Node and Python versions

JGAntunes:
Work for - https://github.com/netlify/pod-workflow/issues/253 - improve the logging of our installed versions as well as the source of said versions (env var, config file, default pinning, etc.)


run-build.sh
```diff
@@ -18,14 +18,13 @@ if [[ ! -d $NETLIFY_REPO_DIR ]]; then
 fi
 cd $NETLIFY_REPO_DIR
```
JGAntunes, line 20: The changes within this file would need to reflected in `buildbot`.

run-build.sh
```diff
@@ -18,14 +18,13 @@ if [[ ! -d $NETLIFY_REPO_DIR ]]; then
...
 
-: ${NODE_VERSION="16"}
-: ${RUBY_VERSION="2.7.2"}
-: ${YARN_VERSION="1.22.10"}
-: ${GO_VERSION="1.16.5"}
-: ${PYTHON_VERSION="3.8"}
```
JGAntunes, line 25: Go and Python defaults have no pinned defaults coming from `buildbot`.  There's little value keeping them here and it only makes things more confusing. Moved it to `run_build_functions` together with `PHP` and `Swift`

kitop:
Closing as stale.

PR #619, opened 2021-09-10
chore(tests): add bats tests (xenial)

JGAntunes:
## Overview

Part of https://github.com/netlify/pod-workflow/issues/137, the idea is to add a set of [bats tests](https://github.com/bats-core/bats-core) with the proposed structure.

```sh
tests/
├─ node/
│  ├─ base.bats
│  ├─ yarn.bats
│  ├─ fixtures/
├─ ruby/
│  ├─ base.bats
│  ├─ fixtures/
├─ <some-other-tool-or-runtime>/
├─ helpers.sh
├─ some-generic-global-tests.bats
├─ fixtures/
```

This is far from set in stone FYI 😅 open to ideas. We can also revisit this as we add more tests and see the need for a different re-org.

Out of scope for this PR are:
- Documentation
- Integration with our CI system.
☝️ to be addressed in follow up PRs.

**Important:** With the addition of an extra stage for our dockerfile, any `docker build` execution aiming at having a prod ready image needs to `--target build-image` specifically (hence all the chaneges to the `Jenkinsfile` and `.workflows`.

## Execution

Currently, these can be executed via:
```sh
docker build -t build-image-xenial-with-tests .
docker run --rm build-image-xenial-with-tests
```

Rendering the following:
```sh
> @netlify/build-image@3.9.0 test /opt/buildhome/test-env
> bats --recursive --timing --tap tests

1..4
ok 1 node version 12.18.0 is installed and available at startup in 18ms
ok 2 grunt-cli is installed and available at startup in 416ms
ok 3 bower is installed and available at startup in 761ms
ok 4 run_yarn sets up new yarn version if different from the one installed, installs deps and creates cache dir in 11007ms
```

Failing assertions produce something like:
```sh
1..4
not ok 1 node version 12.18.1 is installed and available at startup in 71ms
# (from function `assert_output' in file tests/node/../../node_modules/bats-assert/src/assert.bash, line 247,
#  in test file tests/node/base.bats, line 16)
#   `assert_output --partial $NODE_VERSION' failed
#
# -- output does not contain substring --
# substring : 12.18.1
# output    : v12.18.0
# --
#
ok 2 grunt-cli is installed and available at startup in 466ms
ok 3 bower is installed and available at startup in 743ms
ok 4 run_yarn sets up new yarn version if different from the one installed, installs deps and creates cache dir in 9463ms
npm ERR! Test failed.  See above for more details.
```

We could probably benefit from scripts that volumed both the  `tests` dir as well as the `run-build*` scripts in order to provide a better dev ex. IMO we could tackle that afterwards (e.g. as part of the documentation work).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build Docker image
-        run: docker build . --build-arg NF_IMAGE_VERSION=$GITHUB_SHA -t netlify/build:$GITHUB_SHA
+        run: docker build . --build-arg NF_IMAGE_VERSION=$GITHUB_SHA  --target build-image -t netlify/build:$GITHUB_SHA
       - uses: vbrown608/whalescale@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,7 @@ pipeline {
         anyOf { branch 'staging' ; branch 'xenial' ; branch 'trusty' ; buildingTag() }
       }
       steps {
-        sh "docker build --build-arg NF_IMAGE_VERSION=${env.GIT_COMMIT} --build-arg NF_IMAGE_TAG=${env.BRANCH_NAME} -t netlify/build:${env.BRANCH_NAME} -t netlify/build:${env.GIT_COMMIT} ."
+        sh "docker build --build-arg NF_IMAGE_VERSION=${env.GIT_COMMIT} --build-arg NF_IMAGE_TAG=${env.BRANCH_NAME} -t netlify/build:${env.BRANCH_NAME} -t netlify/build:${env.GIT_COMMIT} --target build-image ."
       }
     }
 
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build Docker image
-        run: docker build . --build-arg NF_IMAGE_VERSION=$GITHUB_SHA -t netlify/build:$GITHUB_SHA
+        run: docker build . --build-arg NF_IMAGE_VERSION=$GITHUB_SHA  --target build-image -t netlify/build:$GITHUB_SHA
```
ingride, line 11: this will only run on one build image since xenial and focal are on separate branches. why do we need a target?

ingride (commented):
looks good otherwise

.github/workflows/ci.yml
ingride: oh nvm i see because test is a different stage and we wanna include it in build.

Jenkinsfile
```diff
@@ -25,7 +25,7 @@ pipeline {
         anyOf { buildingTag() }
```
ingride, line 25: should we rename this stage `Build & Test images`?

JGAntunes: I was leaving the test phase out of CI on purpose (as per https://github.com/netlify/pod-workflow/issues/137#issuecomment-917057034 as well as https://github.com/netlify/pod-workflow/issues/251) , so this stage is still only building the actual `Build Image`. I was hoping we could focus on bringing these tests to the CI env on a separate PR?

ingride (approved):
🧪

PR #631, opened 2021-09-17
chore(tests): add bats tests (focal)

JGAntunes:
Same as - https://github.com/netlify/build-image/pull/619 - but adjusted to the `focal` branch.

Related with https://github.com/netlify/pod-workflow/issues/137.

See https://github.com/netlify/build-image/pull/619#issue-993439992 for better context.

PR #632, opened 2021-09-20
chore: add Makefile and docs around our dev/test processes

JGAntunes:
Part of https://github.com/netlify/pod-workflow/issues/137 and a follow up of #631.

This PR adds a `Makefile` as a way to improve local dev experience as well as docs on our dev/test processes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,6 +9,26 @@ please read the [code of conduct](CODE_OF_CONDUCT.md).
 2. Wait for the Jenkins build to finish
 3. Create a PR in the [buildbot](https://github.com/netlify/buildbot) to bump [the version](https://github.com/netlify/buildbot/blob/ddbb47739f5b85c954aad9dc3823ab0676432957/Jenkinsfile#L35) of the `build-image`.
 
+## Development
+
+### Linting
+- [ShellCheck](https://github.com/koalaman/shellcheck) usage is recommended, however it is not enforced.
+
+### Tests
+We have a set of automated tests in [./tests](./tests). These are [bats](https://github.com/bats-core/bats-core) tests that we use to make assertions not only on the correct functioning of our bash/shell scripts, but also of the software provided by our Docker image. Any fix or feature should be accompanied by a set of tests to validate that those changes work as expected. For an overview on how [bats works see here](https://bats-core.readthedocs.io/en/stable/).
+
+### Developing
+
+We provide a [Makefile](./Makefile) with a set of utility targets to help with development.
+
+Some examples:
+
+- `make test` build the test docker image and run the tests in [TAP fromat](http://testanything.org/).
+- `make test-local` creates a volume of the tests directory and the build scripts and run the tests inside the container. Useful when developing locally.
+- `make test-local FILTER=<regex>` provide a filter regex string to your test execution in order to select a specific set of tests.
+- `make run` build the base image and run a bash shell in a container based in it in interactive mode.
+- `make run-local` volume the build scripts and run a bash shell in a container based in the build image.
+
 ## License
 
 By contributing to Netlify's build-image, you agree that your contributions will be licensed
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,9 +12,11 @@ please read the [code of conduct](CODE_OF_CONDUCT.md).
 ## Development
 
 ### Linting
-- [ShellCheck](https://github.com/koalaman/shellcheck) usage is recommended, however it is not enforced.
+
+[ShellCheck](https://github.com/koalaman/shellcheck) usage is recommended, however it is not enforced.
 
 ### Tests
+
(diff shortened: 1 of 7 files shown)


tests/helpers.sh
```diff
@@ -24,3 +24,10 @@ set_fixture_as_repo() {
 setup_tmp_dir() {
   mktemp -d
 }
+
+# Sources nvm.sh, in order for node and npm binaries to be accessible
+source_nvm() {
+  # Disable shellcheck's source file check
+  # shellcheck source=/dev/null
+  source ~/.nvm/nvm.sh
+}
```
JGAntunes, line 33: We now source `nvm` directly in our tests, whenever needed, instead of relying in the Dockerfile CMD/ENTRYPOINT to do such.

CONTRIBUTING.md
```diff
@@ -9,6 +9,28 @@ please read the [code of conduct](CODE_OF_CONDUCT.md).
 2. Wait for the Jenkins build to finish
 3. Create a PR in the [buildbot](https://github.com/netlify/buildbot) to bump [the version](https://github.com/netlify/buildbot/blob/ddbb47739f5b85c954aad9dc3823ab0676432957/Jenkinsfile#L35) of the `build-image`.
 
+## Development
+
+### Linting
+
+[ShellCheck](https://github.com/koalaman/shellcheck) usage is recommended, however it is not enforced.
```
JGAntunes, line 16: I took the liberty of adding this as I consider it a useful tool. Let me know if you have a different view on the subject @Benaiah @ehmicky 👍

ehmicky: Sounds good!

Dockerfile
```diff
@@ -491,6 +491,11 @@ ADD --chown=buildbot package.json /opt/buildhome/test-env/package.json
...
     ln -s /opt/build-bin/build /opt/buildhome/test-env/run-build.sh
+
 ADD --chown=buildbot tests /opt/buildhome/test-env/tests
 WORKDIR /opt/buildhome/test-env
-CMD . ~/.nvm/nvm.sh && npm test
+
+# Set `bats` as entrypoint
+ENTRYPOINT ["node_modules/.bin/bats"]
+# Set the default flags for `bats`
+CMD ["--recursive", "--timing", "--tap", "tests"]
```
ehmicky, line 501: Would switching from `CMD` to `ENTRYPOINT + CMD` be a problem when importing inside the buildbot, since it [overrides only `CMD`](https://github.com/netlify/buildbot/blob/1ec6dc389c9f413d6279ea44d1f11466e1123151/Dockerfile#L60)? Wouldn't Docker interpret it then as "overwrite the argument to `bats`"?

Dockerfile
```diff
@@ -491,6 +491,11 @@ ADD --chown=buildbot package.json /opt/buildhome/test-env/package.json
 RUN cd /opt/buildhome/test-env && . ~/.nvm/nvm.sh && npm i --legacy-peer-deps &&\
     ln -s /opt/build-bin/run-build-functions.sh /opt/buildhome/test-env/run-build-functions.sh &&\
     ln -s /opt/build-bin/build /opt/buildhome/test-env/run-build.sh
+
 ADD --chown=buildbot tests /opt/buildhome/test-env/tests
 WORKDIR /opt/buildhome/test-env
-CMD . ~/.nvm/nvm.sh && npm test
+
+# Set `bats` as entrypoint
+ENTRYPOINT ["node_modules/.bin/bats"]
```
ehmicky, line 499: What are your thoughts on using `["npx", "bats"]` instead?

ehmicky:
Looks good to me! Just added a few questions.

Dockerfile
JGAntunes: This will only apply to our [`test` image](https://github.com/netlify/build-image/blob/focal/Dockerfile#L476-L482), which is never used in production (see [here](https://github.com/netlify/build-image/blob/focal/Jenkinsfile#L28) and [here](https://github.com/netlify/build-image/blob/focal/Jenkinsfile#L19), notice the `--target` flag pointing to [build-image](https://github.com/netlify/build-image/blob/focal/Dockerfile#L1). This means that in these contexts the last test layer is actually never built nor used.)

Dockerfile
JGAntunes: If I understand it correctly, `npx` would still force the sourcing of `nvm.sh`. This means effectively changing the environment we use to run our tests. I'm just concerned that by doing so we might be "hiding" some problem that our test env could/should spot?

Dockerfile
ehmicky: Oh, right!

Dockerfile
ehmicky: Ah, I see. Sounds good to me :+1:

ehmicky (approved):

Dockerfile
haaami01: 👍

PR #637, opened 2021-09-27
docs: Specify minimum Ruby version supported on Focal

verythorough:
Our toolchain on Focal doesn't support versions of Ruby older than 2.4.0. This updates the included software doc to reflect that minimum when describing which versions can be installed via `rvm`.

Adding @KyleBlankRollins for writing review. I opted for `<=` for brevity, but perhaps we should use human language instead?

### DIFF
--- a/included_software.md
+++ b/included_software.md
@@ -11,7 +11,7 @@ The specific patch versions included will depend on when the image was last buil
 * Ruby - `RUBY_VERSION`, `.ruby-version`
   * 2.7.2 (default)
   * 2.6.6
-  * Any version that `rvm` can install.
+  * Any version >= 2.4.0 that `rvm` can install.
 * Node.js - `NODE_VERSION`, `.nvmrc`, `.node-version`
   * 16 (default)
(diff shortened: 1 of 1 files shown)


ehmicky (dismissed):

included_software.md
```diff
@@ -11,7 +11,7 @@ The specific patch versions included will depend on when the image was last buil
 * Ruby - `RUBY_VERSION`, `.ruby-version`
   * 2.7.2 (default)
   * 2.6.6
-  * Any version that `rvm` can install.
+  * Any version >= 2.4.0 that `rvm` can install.
```
KyleBlankRollins: @verythorough, I do like the brevity here and think it's pretty clear.

But, if we're looking for consistency, we try to use "earlier" and "later" wording with versions. So, something like the following may be appropriate:
```suggestion
  * Any version 2.4.0 or later that `rvm` can install.
```

JGAntunes (approved):

PR #647, opened 2021-10-12
docs: remove Trusty included-software link

verythorough:
Just notice that we still link to the Trusty included-software list, so I removed it. I also moved Focal to the top, since it's the default.

We'll also want to copy this change to Xenial. Should I make a separate PR, or shall we just cherry pick it over?

A tahr on a hill. Similar to a goat with long hair like a lion's mane
_A very majestic (and probably trusty) tahr_

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,8 @@ Netlify maintains multiple build images for testing new development as well as s
 
 The following images are currently available:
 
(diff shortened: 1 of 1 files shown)


ehmicky (approved):
Usually we make a separate PR.
However, I am now wondering whether this is worth double reviews, when we could just cherry-picked for simple cases like this? :thinking:

PR #659, opened 2021-10-25
fix: buildbot group xenial

JGAntunes:
🎉 Thanks for submitting a pull request! 🎉

#### Summary

Same as #652, ported to our Xenial branch.

Note, piepline is failing because of #660 

---

For us to review and ship your PR efficiently, please perform the following steps:

- [x] Open a [bug/issue](https://github.com/netlify/build-image/issues/new/choose) before writing your code 🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [x] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and passes our tests.
- [x] Update or add tests (if any source code was changed or added) 🧪
- [x] Update the [included software doc](../included_software.md) (if you updated included software) 📄
- [x] Update or add documentation (if features were changed or added) 📝
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**

PR #653, opened 2021-10-19
chore(cci): fix git tag pipeline

JGAntunes:
This PR fixes some issues with the git tag specific part of the CCI pipeline. I've commented directly on the changes to better explain some of what happened. Meanwhile the pipeline executions:
- For tags: https://app.circleci.com/pipelines/github/netlify/build-image/71/workflows/717ac4cd-5d4e-4461-8610-e63c21e2bb47
- For branches: https://app.circleci.com/pipelines/github/netlify/build-image/70/workflows/800cf49c-71c1-49d2-8970-97705c89f735

Work as expected and the images are correctly pushed to the repo: https://hub.docker.com/r/jgantunesntf/netlify-build/tags?page=1&ordering=last_updated

Inspecting the image layers we have:

```bash
$docker image inspect jgantunesntf/netlify-build:v-test-circleci-2
(...)
        "Id": "sha256:c2a0c724ec73d0fed230129b78e90dc641edb18e2cd3741aa395c33d6124e46c",
        "RepoTags": [
            "jgantunesntf/netlify-build:v-test-circleci-2"
        ],
(...)
            "Env": [
                (..)
                "NF_IMAGE_VERSION=ba6536d755dc20f1a482499fe6f3cb6accdc6598",
                "NF_IMAGE_TAG=v-test-circleci-2"
            ],
(...)
        "RootFS": {
            "Type": "layers",
            "Layers": [
                "sha256:da55b45d310bb8096103c29ff01038a6d6af74e14e3b67d1cd488c3ab03f5f0d",
                "sha256:108ce1f087f59a8cbb0b41e9c8d5feaf8769838f74f35180f23ed7167e3e480d",
                "sha256:202c3e28dd9ee68d523e75fd0a9cb0a18041d5792ee9a96147edb6877fcd1b87",
                "sha256:ee7bdded6b14e200fa2870a9d7c2bac81ff7240f82391da0c8166a426462d539",
                "sha256:220188ddd851224286eb1707e05df6860995c90be672131c23182f0a0945a783",
                "sha256:f95f9daa53f04cbf38d4f4f4b72e59b8cfb6496a464f6405bc982bad0c993f1d",
                "sha256:650ba2111df48a4a520a7e45b3881623f4267ab18968e7c71b9dc156b615bd2a",
                "sha256:002adc2c073f7674ec901ce77d1530e49fb1659bcb755fc592336397388c6ed3",
[...]

PR #651, opened 2021-10-15
chore: CircleCI project setup

JGAntunes:
This PR adds support for CCI as a replacemente for our current Jenkins pipeline. ~~I've commented out the Jenkins pipelines for now (in a similar strategy as we've done within buildbot) and we're currently skipping the pipeline altogether~~ (**Edit**: opted for actually keeping the pipeline active for now (see next steps below).

Besides replacing the pipeline we had this also adds a couple of new things:
- **Builds out our test image** (based in the "production" one)
- **Runs our newly added tests** (#631 #619 and https://github.com/netlify/pod-workflow/issues/137) within the CI pipeline.
- **Always pushes the newly created image tags** to the registry for trusted sources (branches from folks with push access) and, **pending on approval** from someone with push access, **also pushes for PRs from forked repos**.

In order to speed things up we build all the images within a single job `build-and-push-images`. These images are then saved via a `docker save` and persisted locally for later use within the workflow (and in order to avoid having to rebuild the whole thing for testing purposes or pushing tags (in forks). Although CCI has docker layer caching, this has its nuances and doesn't always kick in, specially for jobs within different layers of the workflow).

**NOTE:** For testing purposes I'm still pusing images to a personal repo I've created.

Next steps (**edited**):
- Merge this PR
- Validate the flow works accordingly for
  - Tags
  - Forked repos
- Replace the repo we push to and comment out the Jenkins pipeline

Other relevant resources:
- [The docker repo I've used](https://hub.docker.com/repository/docker/jgantunesntf/netlify-build)
- [The multiple pipeline executions](https://app.circleci.com/pipelines/github/netlify/build-image?branch=circleci-project-setup)
-  [A test of the tag flow](https://app.circleci.com/pipelines/github/netlify/build-image/48/workflows/60d5972d-3467-44a8-bd1c-09c3e7cc9dda) (which squashes images) I've ran for a branch (commented out the filters)

For us to review and ship your PR efficiently, please perform the following steps:

- [x] Open a [bug/issue](https://github.com/netlify/build-image/issues/new/choose) before writing your code 🧑‍💻. This ensures
      we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or
      something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [x] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [ ] ~~Update or add tests (if any source code was changed or added) 🧪~~
[...]

PR #660, opened 2021-10-25
chore: cci xenial pipeline

JGAntunes:
#### Summary

This PR picks up on both #651 and #653 and ports them over to `Xenial` (literally just a rebase). This is required because we still merge PRs against this branch (e.g. see https://github.com/netlify/build-image/pull/659). Meaning that, in order for the pipelines to run, we need these to be present in the `base` branch. 

Related with https://github.com/netlify/pod-workflow/issues/251

---

🐱

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -204,7 +204,7 @@ RUN \
   tar xvf vips-8.6.2.tar.gz && \
   cd libvips-8.6.2 && \
   ./autogen.sh && \
-  ./configure --enable-debug=no --enable-docs=no --without-python --without-orc --without-fftw --without-gsf $1 && \
+  ./configure --enable-debug=no --enable-docs=no --without-python --without-orc --without-fftw --without-gsf && \
   make && \
   make install && \
   ldconfig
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -204,7 +204,7 @@ RUN \
   tar xvf vips-8.6.2.tar.gz && \
   cd libvips-8.6.2 && \
   ./autogen.sh && \
-  ./configure --enable-debug=no --enable-docs=no --without-python --without-orc --without-fftw --without-gsf $1 && \
+  ./configure --enable-debug=no --enable-docs=no --without-python --without-orc --without-fftw --without-gsf && \
```
JGAntunes, line 207: This was triggering our linting:
https://app.circleci.com/pipelines/github/netlify/build-image/84/workflows/3aff0a58-e9f7-44ee-bb35-cd511c7d4df9/jobs/199

Seems like it was introduced with the original commit for this change and stayed that way ever since - https://github.com/netlify/build-image/commit/68bf2c7baa750ffdc77a33493ceaa19a7765ad65

My understanding is that this `$1` produces nothing at all, given it's not a valid `ARG` or `ENV` var that we pass during our build execution.

ehmicky: Yes, I don't think `$1` is valid within `Dockerfile`.
Looking at the `libvips` docs, it does not seem like `./configure` needs any argument.

ehmicky (approved):

PR #669, opened 2021-10-26
chore(ci): drop Jenkins and move to the correct docker repo (xenial)

JGAntunes:
🎉 Thanks for submitting a pull request! 🎉

#### Summary

Same as #668 but for the `xenial` branch. Related with netlify/pod-workflow#251 this PR comments out our Jenkinsfile and starts pushing images from circle ci to our docker repo.

---

For us to review and ship your PR efficiently, please perform the following steps:

- [ ] Open a [bug/issue](https://github.com/netlify/build-image/issues/new/choose) before writing your code 🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [ ] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and passes our tests.
- [ ] Update or add tests (if any source code was changed or added) 🧪
- [ ] Update the [included software doc](../included_software.md) (if you updated included software) 📄
- [ ] Update or add documentation (if features were changed or added) 📝
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**

🦋

PR #668, opened 2021-10-26
chore(ci): drop Jenkins and move to the correct docker repo

JGAntunes:
🎉 Thanks for submitting a pull request! 🎉

#### Summary
Related with https://github.com/netlify/pod-workflow/issues/251 this PR comments out our Jenkinsfile and starts pushing images from circle ci to our docker repo.

---

For us to review and ship your PR efficiently, please perform the following steps:

- [x] Open a [bug/issue](https://github.com/netlify/build-image/issues/new/choose) before writing your code 🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [x] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and passes our tests.
- [x] Update or add tests (if any source code was changed or added) 🧪
- [x] Update the [included software doc](../included_software.md) (if you updated included software) 📄
- [x] Update or add documentation (if features were changed or added) 📝
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**

🦆

PR #670, opened 2021-10-26
chore(ci): clean whalescale github action

JGAntunes:

#### Summary

Related with https://github.com/netlify/pod-workflow/issues/251. This PR cleans our [whalescale GitHub action](https://github.com/netlify/whalescale). We renamed it, so that is more obvious to anynone what it's doing and we also make use of a docker-layer caching action in order to improve its speed. The alternative to actually building the image would be to wait for Circle's check to finish and pull the image (via something like https://github.com/marketplace/actions/wait-for-check) however this would likely be slower than just building the image and we would still be actively running an action on the background (it just polls GitHub's checks API).

---

🦤


ehmicky:
This is probably unrelated to this PR, but `docker push` in CI is failing with an authentication error: is this expected?

Also, the new Whalescale GitHub action is timing out. This might also be expected?

JGAntunes:
> This is probably unrelated to this PR, but `docker push` in CI is failing with an authentication error: is this expected?

Yeah, this is because of #669 and #668. We've just changed the Circle CI secrets so until those PRs are merged these checks will fail.

> Also, the new Whalescale GitHub action is timing out. This might also be expected?

Seems like it went alright - https://github.com/netlify/build-image/actions/runs/1386925225?

ingride (approved):
🚤

PR #705, opened 2021-12-12
Brew install before Rubygems, Pip etc.

### LINKED ISSUE #522: Run homebrew install before other installs
We can use Homebrew to install system libraries that may be required by Python/Ruby/Node packages. For that reason, we should run the Homebrew installation step before installing other languages' packages.

ryantownsend:
#### Summary

Fixes #522

Currently, Homebrew is only executed after Rubygems/Pip dependencies have been installed – this means any packages which build using libraries which aren't native to the build image fail.

#### A picture of a cute animal

My girlfriend's cat, Ivan


kitop:

PR #710, opened 2021-12-13
fix: set `NETLIFY=true` at the build-image level

JGAntunes:

#### Summary

Fixes https://github.com/netlify/buildbot/issues/1892. This PR changes the `NETLIFY=true` flag to be exposed at the `build-image` build level, meaning it will be set for all of Buildbot's executions, including calls to the ignore script.

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -480,6 +480,8 @@ ENV NF_IMAGE_TAG ${NF_IMAGE_TAG:-latest}
 ARG NF_IMAGE_NAME
 ENV NF_IMAGE_NAME ${NF_IMAGE_NAME:-focal}
 
+# CI signal, we set it at image level so that any Buildbot command will inherit this flag
+ENV NETLIFY=true
 
 ################################################################################
 #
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -481,6 +481,8 @@ ENV NF_IMAGE_TAG ${NF_IMAGE_TAG:-latest}
 ARG NF_IMAGE_NAME
 ENV NF_IMAGE_NAME ${NF_IMAGE_NAME:-focal}
 
+# CI signal, we set it at image level so that any Buildbot command will inherit this flag
+ENV NETLIFY=true
```
ehmicky, line 485: I believe this should work, since the `buildbot` Dockerfile will run this instruction and `ENV` variables are used when the Docker container is run, and `@netlify/build` is a child process of the process run by that container.

However we should double-check in the buildbot PR upgrading the `build-image` that the `NETLIFY` environment variable is set to `true` inside the build command, just to be sure.

ehmicky (approved):

Dockerfile
JGAntunes: Yep agree 👍 in fact I plan to test out a pre-release here, run this through a buildbot release without actually releasing a new `build-image`. I'll update this PR once I have some results.

lukasholzer (approved):

lukasholzer:
@JGAntunes and this one as well? 😅

JGAntunes:
Closing as per - https://github.com/netlify/build-image/pull/711#issuecomment-1170273858

PR #713, opened 2021-12-16
fix: Bump Go to 1.17 and omit patch version

akahn:
#### Summary

1. Bump to latest Go minor version
2. Omit the patch version. By omitting the patch version, the latest 1.7.x will be used, which helps ensure that security fixes are brought in automatically.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -402,7 +402,7 @@ ENV PATH "$PATH:/opt/buildhome/.gimme/bin"
 ENV GOPATH "/opt/buildhome/.gimme_cache/gopath"
 ENV GOCACHE "/opt/buildhome/.gimme_cache/gocache"
 # Install the default version
-ENV GIMME_GO_VERSION "1.17"
+ENV GIMME_GO_VERSION "1.17.x"
 ENV GIMME_ENV_PREFIX "/opt/buildhome/.gimme/env"
 RUN gimme | bash
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -402,7 +402,7 @@ ENV PATH "$PATH:/opt/buildhome/.gimme/bin"
 ENV GOPATH "/opt/buildhome/.gimme_cache/gopath"
 ENV GOCACHE "/opt/buildhome/.gimme_cache/gocache"
 # Install the default version
-ENV GIMME_GO_VERSION "1.16.5"
+ENV GIMME_GO_VERSION "1.7"
 ENV GIMME_ENV_PREFIX "/opt/buildhome/.gimme/env"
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -402,7 +402,7 @@ ENV PATH "$PATH:/opt/buildhome/.gimme/bin"
 ENV GOPATH "/opt/buildhome/.gimme_cache/gopath"
 ENV GOCACHE "/opt/buildhome/.gimme_cache/gocache"
 # Install the default version
-ENV GIMME_GO_VERSION "1.16.5"
+ENV GIMME_GO_VERSION "1.7"
```
mraerino: _did you mean `1.17`?_

JGAntunes:
Hey @akahn 👋 so, the way default version are currently working is a bit... complicated. See - https://github.com/netlify/pod-workflow/issues/253#issuecomment-916205113 - essentially, the default version set here will have little impact into what's running on the customer side given to the pinning taking place at the site level within Bitballoon. The current strategy is to usually have a match between what is being pinned by default now and what we install by default within the `build-image` (although it doesn't seem to be the case for Go 🤔), until we have a chance to actually address this problem as a whole.

Dockerfile
akahn: Yes, thank you.

run-build-functions.sh
```diff
@@ -655,13 +655,14 @@ install_dependencies() {
...
     if [ $? -eq 0 ]
     then
-      source $HOME/.gimme_cache/env/go$installGoVersion.linux.amd64.env
+      source $HOME/.gimme_cache/env/go$resolvedGoVersion.linux.amd64.env
     else
-      echo "Failed to install Go version '$installGoVersion'"
+      echo "Failed to install Go version '$resolvedGoVersion'"
       exit 1
     fi
   else
```
JGAntunes: Wondering if we could add a test case for this @akahn? (i.e. https://github.com/netlify/build-image/blob/focal/tests/node/base.bats#L13-L18). For more info around our tests see - https://github.com/netlify/build-image/blob/focal/CONTRIBUTING.md#tests

JGAntunes (dismissed):
Left a suggestion in order to improve our test coverage in this area @akahn 👍

Also, we have an automated CHANGELOG being generated based on the `release-please` process (which itself relies on commit messages following the conventional commit format) - https://github.com/netlify/build-image/blob/focal/CONTRIBUTING.md#developing. You might need to update the PR title/commit msg (for single commit PRs).

aitkenster (approved):

akahn:
👏 👏 👏

PR #725, opened 2022-01-19
chore: warn when NODE_ENV == "production"

### LINKED ISSUE #575: Print a warning message when `NODE_ENV` is `production`
When `NODE_ENV` is `production`, `devDependencies` are not installed.
This can lead to situations where the build fails due to missing dependencies, which results in bug reports and support tickets.

Conceptually, it seems to me builds should never use `NODE_ENV="production"`. Are there any legitimate use cases for this? It seems like most builds would need development dependencies and are technically part of the development flow, not production.

What are your thoughts on printing a warning message when `NODE_ENV` is `production`, warning that development dependencies will not be installed and might be missing?

cc @kaganjd

aitkenster:
#### Summary

Fixes #575

---

image

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -281,11 +281,6 @@ install_dependencies() {
     fi
   fi
 
-  if [ "$NODE_ENV" == "production" ]
-  then
-     warn "NODE_ENV is set to 'production'. Any devDependencies in package.json will not be installed"
-  fi
-
   # Automatically installed Build plugins
   if [ ! -d "$PWD/.netlify" ]
   then
@@ -511,6 +506,11 @@ install_dependencies() {
 
   if [ -f package.json ]
   then
+    if [ "$NODE_ENV" == "production" ]
+    then
+      warn "The environment variable 'NODE_ENV' is set to 'production'. Any 'devDependencies' in package.json will not be installed"
+    fi
+
     if [ "$NETLIFY_USE_YARN" = "true" ] || ([ "$NETLIFY_USE_YARN" != "false" ] && [ -f yarn.lock ])
     then
       run_yarn $YARN_VERSION
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -281,6 +281,11 @@ install_dependencies() {
     fi
   fi
 
+  if [ "$NODE_ENV" == "production" ]
+  then
+     warn "NODE_ENV is set to 'production'. Any devDependencies in package.json will not be installed"
+  fi
+
   # Automatically installed Build plugins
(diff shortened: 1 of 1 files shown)


run-build-functions.sh
```diff
@@ -281,6 +281,11 @@ install_dependencies() {
     fi
   fi
 
+  if [ "$NODE_ENV" == "production" ]
+  then
+     warn "NODE_ENV is set to 'production'. Any devDependencies in package.json will not be installed"
```
ehmicky: What are your thoughts on the following?

```suggestion
     warn "The environment variable 'NODE_ENV' is set to 'production'. Any 'devDependencies' in package.json will not be installed."
```

run-build-functions.sh
```diff
@@ -888,3 +893,7 @@ unset_go_import_path() {
     unlink $GOPATH/src/$GO_IMPORT_PATH
   fi
 }
+
+warn() {
+  echo "WARNING: $1"
+}
```
ehmicky: [sand] Some IDEs automatically insert a final newline to each file to our [`.editorconfig` file](https://github.com/netlify/build-image/blob/bac1417dd652181d67060aaf756418e6d8ca485a/.editorconfig#L11).
I.e. an unrelated PR might add a newline there.

run-build-functions.sh
```diff
@@ -281,6 +281,11 @@ install_dependencies() {
     fi
   fi
 
+  if [ "$NODE_ENV" == "production" ]
```
ehmicky: Should we put this right before `run_npm`/`run_yarn` instead? Like this, we will not print this when the site has no `package.json`. Granted though that having a `NODE_ENV` environment variable but no `package.json` should be unlikely.

ehmicky:
Thanks @aitkenster!

This looks good to me, except for some small details.

After shipping, we should monitor on Humio how many sites are printing this warning, just in case there happened to be a legitimate reason why some users might need to use `NODE_ENV=production` (which is unclear at the moment).

run-build-functions.sh
aitkenster: Yes, I'll change it 👍 I think it's better without the final "." though to be consistent with the rest of the file.

run-build-functions.sh
ehmicky: Note: link to setups for `.editorconfig` for different IDEs: https://editorconfig.org/#pre-installed

run-build-functions.sh
ehmicky: Good point! :+1: (pun intended)

ehmicky (approved):

PR #750, opened 2022-03-03
feat: add a mirror for when `nodejs.org/dist` is down

### LINKED ISSUE #469: Better error handling of `nvm` and `nodejs.org/dist`
We run `nvm install` to check what the latest matching Node.js version is (e.g. `NODE_VERSION` `12` would match `12.18.3`), and download it unless it is available locally:

https://github.com/netlify/build-image/blob/450c24b893036c2813a0b0d0a2a05bcdd81935a4/run-build-functions.sh#L221

Even if the desired Node.js version is already available locally, we still need to use `nvm install` (and make an HTTP request) to check whether a newer matching Node.js version is available. However, if the command fails because `nodejs.org` is down, we should default to using a local Node.js version instead, providing it matches the desired Node.js version. At the moment, we fail instead, which created an outage on 2020-08-30.

https://github.com/netlify/build-image/blob/450c24b893036c2813a0b0d0a2a05bcdd81935a4/run-build-functions.sh#L233

ehmicky:
Fixes https://github.com/netlify/build-image/issues/469

This adds a mirror website to fallback when https://nodejs.org/dist/ is down, which has happened a couple of times in the past, creating incidents.

Tested successfully in production in https://app.netlify.com/sites/mickael-hebert/deploys/62263574097a570008bb8344#L24 :heavy_check_mark: 

---

### DIFF
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -72,6 +72,31 @@ mkdir -p $NETLIFY_CACHE_DIR/.cargo
 : ${NPM_FLAGS=""}
 : ${BUNDLER_FLAGS=""}
 
+# Run `nvm install ...`
+nvm_install() {
+  if nvm install "$@"; then
+    return 0
+  fi
+
+  # nvm install might fail due to either:
+  #  - A user error, e.g. the NODE_VERSION being an invalid version
+  #  - A third-party outage of nodejs.org/dist
+  # We distinguish between those by making a HEAD HTTP request to nodejs.org/dist
+  if curl --head --fail "$DEFAULT_NODE_MIRROR"; then
+    return 1
+  fi
+
+  echo "Could not connect to $DEFAULT_NODE_MIRROR"
+  echo "Connecting to fallback $FALLBACK_NODE_MIRROR"
+  NVM_NODEJS_ORG_MIRROR="$FALLBACK_NODE_MIRROR" nvm install "$@"
+}
+
+# Environment variable used by nvm to customize where the Node.js binaries are
+# downloaded from. Our default value matches the one from nvm.
+DEFAULT_NODE_MIRROR="${NVM_NODEJS_ORG_MIRROR:-"https://nodejs.org/dist"}"
+# When nodejs.org/dist/ is down, we try to use a mirror site as a fallback.
+FALLBACK_NODE_MIRROR="https://npmmirror.com/mirrors/node"
+
 install_deps() {
   [ -f $1 ] || return 0
   [ -f $3 ] || return 0
@@ -257,7 +282,7 @@ install_dependencies() {
     echo "Attempting node version '$NODE_VERSION' from .node-version"
   fi
 
-  if nvm install --no-progress $NODE_VERSION
+  if nvm_install --no-progress $NODE_VERSION
   then
     NODE_VERSION=$(nvm current)
     # no echo needed because nvm does that for us


run-build-functions.sh
```diff
@@ -72,6 +72,29 @@ mkdir -p $NETLIFY_CACHE_DIR/.cargo
...
+nvm_install() {
+  if nvm install "$@"; then
+    return 0
+  fi
+
+  # nvm install might fail due to either:
+  #  - A user error, e.g. the NODE_VERSION being an invalid version
+  #  - A third-party outage of nodejs.org/dist
+  # We distinguish between those by making a HEAD HTTP request to nodejs.org/dist
+  if curl --head --fail "$DEFAULT_NODE_MIRROR"; then
```
ehmicky, line 85: `nvm` uses `curl` internally, so we re-use it too to be closer to the `nvm` logic.

run-build-functions.sh
```diff
@@ -72,6 +72,29 @@ mkdir -p $NETLIFY_CACHE_DIR/.cargo
...
+    return 1
+  fi
+
+  NVM_NODEJS_ORG_MIRROR="$FALLBACK_NODE_MIRROR" nvm install "$@"
+}
+
+# Environment variable used by nvm to customize where the Node.js binaries are
+# downloaded from. Our default value matches the one from nvm.
+DEFAULT_NODE_MIRROR="${NVM_NODEJS_ORG_MIRROR:-"https://nodejs.org/dist"}"
```
ehmicky, line 96: See https://github.com/nvm-sh/nvm/blob/39d9a42c359df5dbf37c5000af8192f14a280b94/nvm.sh#L1938

run-build-functions.sh
```diff
@@ -72,6 +72,29 @@ mkdir -p $NETLIFY_CACHE_DIR/.cargo
...
+# When nodejs.org/dist/ is down, we try to use a mirror site as a fallback.
+FALLBACK_NODE_MIRROR="https://npmmirror.com/mirrors/node"
```
ehmicky, line 98: This mirror is actually mostly meant to download Node.js from China.
This is the only mirror for `nodejs.org/dist` I could find. It is much slower, but it works. If you know of any others, please add it so we have multiple fallbacks.

run-build-functions.sh
```diff
@@ -72,6 +72,31 @@ mkdir -p $NETLIFY_CACHE_DIR/.cargo
...
+  echo "Could not connect to $DEFAULT_NODE_MIRROR"
+  echo "Connecting to fallback $FALLBACK_NODE_MIRROR"
+  NVM_NODEJS_ORG_MIRROR="$FALLBACK_NODE_MIRROR" nvm install "$@"
+}
+
+# Environment variable used by nvm to customize where the Node.js binaries are
+# downloaded from. Our default value matches the one from nvm.
+DEFAULT_NODE_MIRROR="${NVM_NODEJS_ORG_MIRROR:-"https://nodejs.org/dist"}"
+# When nodejs.org/dist/ is down, we try to use a mirror site as a fallback.
+FALLBACK_NODE_MIRROR="https://npmmirror.com/mirrors/node"
```
erezrokah, line 98: Let's ping @netlify/security to see if they have any concerns using this fallback mirror

erezrokah (changes requested):
Thanks @ehmicky, the code changes look good.

I added a comment to get feedback from our security team

run-build-functions.sh
ehmicky: Sounds good :+1: 
Note to @netlify/security: this mirror would only be used when `nodejs.org` has an outage.

ehmicky: @dustincrogers What do you think?

dustincrogers:
Please excuse the delayed response. 
Im not sure if this is the best practice tbh @ehmicky . Im skeptical about any unvetted mirror. Let me run the risk by the security team and business. My main concern is I dont see anything on that site alluding to their security practices, etc. 

>which has happened a couple of times in the past, creating incidents.

Could we get some ballpark metrics?
- Number of incidents/likelihood
- Average duration
- What is our appetite for downtime before we would need a solution? 30 mins? 4hrs?

ehmicky:
This makes sense, and I believe @erezrokah had similar security concerns.
I believe we've had 2 incidents in the last 2.5 years where https://nodejs.org/dist/ was down for approximately 1-2 hours.
@erezrokah Should I close this PR? I do not believe there are other available Node.js mirrors. We could consider caching Node.js binaries ourselves instead, but that's a different solution which would require a different PR.

erezrokah:
Thanks @ehmicky. I think we can close this and caching sounds like a good approach to explore in another PR

PR #754, opened 2022-03-07
chore: use TARGETARCH arg in Dockerfile

### LINKED ISSUE #715: Can't build image on Apple Silicon
[Some](https://github.com/netlify/build-image/blob/focal/Dockerfile#L303-L305) [places](https://github.com/netlify/build-image/blob/focal/Dockerfile#L194-L203) in the `Dockerfile` hard-code the architecture of installed packages as `amd64`. On an Apple computer with M1 (arm64) architecture, this produces the following error:

```
#6 2.597  package architecture (amd64) does not match system (arm64)
```

We should update these install commands to use one of the [Docker built-in variables](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope), such as `TARGETARCH`. This way, the packages will be installed according to the architecture of the Docker engine. 

There may also be other issues M1 users face after this, the immediate problem, is solved.

aitkenster:

Uses the TARGETARCH arg to allow for builds with apple silicon

#### Summary

Fixes https://github.com/netlify/build-image/issues/715

---

~~Update or add tests (if any source code was changed or added) 🧪~~
~~Update the [included software doc](../included_software.md) (if you updated included software) 📄~~
~~Update or add documentation (if features were changed or added) 📝~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
 FROM ubuntu:20.04 as build-image
+ARG TARGETARCH
+ENV TARGETARCH "${TARGETARCH}"
 
 LABEL maintainer Netlify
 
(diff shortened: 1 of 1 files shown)


lukasholzer (dismissed):
LGTM I think this is needed to support the M1 crew like myself properly ;)

lukasholzer (dismissed):

Dockerfile
```diff
@@ -1,4 +1,6 @@
 FROM ubuntu:20.04 as build-image
+ARG TARGETARCH
+ENV TARGETARCH "${TARGETARCH}"
```
JGAntunes, line 3: Wondering if we could simplify this by having `ARG` have a default value

```
ARG TARGETARCH=amd64
```

And that way avoid setting it in the follow up commands?

JGAntunes (approved):
🚀

PR #756, opened 2022-03-08
chore: use TARGETARCH arg in xenial

aitkenster:

#### Summary

Xenial version of https://github.com/netlify/build-image/tree/nicola/use-targetarch-arg


kitop:
@aitkenster is this still needed? I see it has landed in focal but not in Xenial. With the Xenial deprecation, should we still do this before that happens or is it not worth it?

aitkenster:
I think we can close it

PR #808, opened 2022-07-03
fix: cache ruby version gems dir

aitkenster:

#### Summary

~~(This needs some more testing)~~
Series of tests with various ruby + bundler configurations performed on this site: https://app.netlify.com/sites/visionary-granita-18cccb/deploys

The main problem with the 3.1.2 Ruby version seems to be that there is now a mismatch between the contents of two directories: `./rvm/ruby-$VERSION/gems` and `/ruby-$VERSION/rubies`. Until now, we have only cached `/rubies`, but if we also cache `/gems` then as long as the uncached build is successful, no matter which ruby version we are using, both folders should contain the right files and the cached build should also succeed.

Relates to https://github.com/netlify/pillar-workflow/issues/558

---

image


ingride (approved):
PR seems to suggest it needs more testing. can we post the results of that in the description? 

🚢 if tests are 🍏

PR #813, opened 2022-07-07
chore: Fix tiny issues in CONTRIBUTING.md

danez:
#### Summary

n/t


JGAntunes (approved):
📄

danez:
@JGAntunes @kitop Is it expected that circleCI is not triggered when kodiak is merging in the default branch?

kitop:
> @JGAntunes @kitop Is it expected that circleCI is not triggered when kodiak is merging in the default branch?

It is not. Something was off with the webhook, got a 400 response for that. I just re-triggered and now it's building.

danez:
I wonder if it worked now because YOU triggered the webhook and you are properly authorized in circle. But maybe kodiak and other integrations do not have the correct access. Because I have seen the exact same behavior before in this repo in another PR of mine.

kitop:
Oh, interesting! The webhook was sent with the exact same payload, so that's weird. But we've seen it happen with the tags the other day too. I'll try to reach out to CircleCI support to understand what could be going on.

PR #814, opened 2022-07-07
feat: add musl libc and its tools to build image

danez:
#### Summary

We had a [customer report](https://answers.netlify.com/t/building-ring-crate-for-rust-based-function/64265) that they weren't able to have their rust function compiled in our build pipeline. As it turned out they use the crate `ring` which requires `musl-gcc` to compile itself. 

Now the obvious fix would be to have musl dependencies available (which is what I am doing in this PR). I'm not sure though if we actually want to have musl available in the build-image? To me it seems to make sense to have the musl-tools in the build environment, because we [tell people to use musl](https://www.netlify.com/blog/2021/10/14/write-netlify-functions-in-rust/#deployment) and also [do it by default](https://github.com/netlify/zip-it-and-ship-it/blob/main/src/runtimes/rust/constants.ts#L1). In theory glibc would work as well, as[ we use Amazon Linux 2](https://github.com/netlify/bitballoon/blob/00a35db65833350530f4092ef000362572db1608/app/models/cloud_function/aws_lambda.rb#L22), so glibc [is new enough for rust](https://github.com/awslabs/aws-lambda-rust-runtime/discussions/306#discussioncomment-485478).

I wasn't able to really reproduce to compile `ring`, because rust is segfaulting when I try to run it inside the build image (amd64) on my m1 host.

Do I need to include musl/musl-tools in the included_software.md? I don't see gcc there either so I wasn't sure if musl-tools would count as a tool.

The new additional packages are: `musl, musl-dev, musl-tools`

Fixes netlify/pillar-runtime#401

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -147,6 +147,10 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         libvips-tools \
         libyaml-dev \
         mercurial \
+        # musl and musl-tools are needed for certain rust dependencies (ring) to compile correctly
+        # see https://github.com/netlify/pillar-runtime/issues/401
+        musl \
+        musl-tools \
         nasm \
         openjdk-8-jdk \
         optipng \
--- a/focal.yaml
+++ b/focal.yaml
@@ -150,3 +150,9 @@ commandTests:
     args: ["--version"]
     expectedOutput:
       - rustup 1\.\d+\.\d+
+
+  - name: "Musl library and its tools are available"
+    command: "musl-gcc"
+    args: ["--version"]
+    expectedOutput:
+      - Ubuntu 9\.4\.\d+ # version of gcc, as musl-gcc is just a wrapper


focal.yaml
```diff
@@ -54,3 +54,8 @@ commandTests:
     command: "dotnet"
     args: ["--version"]
     expectedOutput: ["6.0"]
+
+  - name: "Musl library and its tools are available"
+    command: "musl-gcc"
+    args: ["--version"]
+    exitCode: 0
```
lukasholzer: can you please assert the version in the output as well? (you can use regexp as well)

lukasholzer (changes requested):

Dockerfile
```diff
@@ -137,6 +137,8 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         libvips-tools \
         libyaml-dev \
         mercurial \
+        musl \
```
lukasholzer, line 152: can you add a comment to the two lines that it's for rust?

like:

```
# used for rust see issue #issueNumber
musl \ 
musl-tools \
```

Trying to get the use cases for tools (as we have probably a lot that we don't need)

lukasholzer: ( a better longer version of the comment is appreciated as well )

focal.yaml
danez: Yes, it is the gcc version, but still makes sense.

Dockerfile
danez: I guess dockerlint does not like my comment. 
https://github.com/RedCoolBeans/dockerlint/issues/54
https://github.com/RedCoolBeans/dockerlint/issues/73

Might make sense to look into using hadolint instead, as dockerlint seems unmaintained.

lukasholzer (dismissed):
Thx a lot 👍

lukasholzer:
have you created a PR for xenial as well?

lukasholzer:
would love to get @JGAntunes opinion here if we should do this in xenial as well :)

Dockerfile
danez: @lukasholzer What should I do about this here?

danez:
waiting for #818 and hadolint, and then rebase/merge and backport to xenial.

lukasholzer (approved):

PR #828, opened 2022-08-11
chore: build image for multiple CPU architectures linux/amd64,linux/arm64

### LINKED ISSUE #821: Container-structure-tests fail for arm64
When building the new Dockerfile on Apple Silicon for arm64 (with the changes from @lukasholzer), the container-structure-tests are failing now:

```
=== RUN: Command Test: hugo
--- FAIL
duration: 289.941792ms
Error: Error creating container: API error (400): failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "hugo": executable file not found in $PATH: unknown
=== RUN: Command Test: Golang
--- FAIL
duration: 19.521397542s
stderr: /opt/build-bin/run-build-functions.sh: line 912: /opt/buildhome/.gimme/env/go1.18.linux.amd64.env: No such file or directory
bash: line 2: go: command not found

Error: Expected string 'go version go1.18 linux/amd64' not found in output ''
Error: Test 'Golang' exited with incorrect error code. Expected: 0, Actual: 127
--- FAIL
duration: 0s
Error: /opt/build-bin/run-build-functions.sh has incorrect permissions. Expected: -rwxrwxr-x, Actual: -rwxr-xr-x
=== RUN: Metadata Test
--- FAIL
duration: 0s
Error: env var TARGETARCH value arm64 does not match expected value: amd64
=== RUN: Command Test: Homebrew
--- FAIL
duration: 295.3165ms
Error: Error creating container: API error (400): failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "brew": executable file not found in $PATH: unknown
```

### LINKED ISSUE #822: Create multiarch images and push them to the dockerhub
As in #818 we introduced the arm64 compatibility for the image now we need to create a multi arch image and push it:

we need to change our pipeline to run with buildx

As far as I've seen the regular docker orb does not provide a way for it. Basically we need to do this:

```bash
# use the buildx driver
docker buildx create --use

docker buildx build \
  --tag=docker.io/<< parameters.image-name >>:${SHA1_IMAGE_TAG} \
  --tag=docker.io/<< parameters.image-name >>:${BRANCH_IMAGE_TAG} \
  --build-arg NF_IMAGE_VERSION=${CIRCLE_SHA1} \
  --build-arg NF_IMAGE_TAG=${VERSION_TAG} \
  --target << parameters.build-target >> <<# parameters.squash >> --squash <</ parameters.squash >> \
  --platform linux/amd64,linux/arm64 \
  --push \
  .
```

This should then cross compile the image in parallel and build for both platforms once this is pushed.

CleanShot 2022-07-15 at 12 57 47@2x

This is the prerequisite to get the buildbot image working for Mac M1 arm64 as well.

lukasholzer:

#### Summary

Fixes #822 
Fixes #821

- Creates an image for each CPU architecture (amd64, arm64) and pushes it to the registry
- Combines those images to a multiarch image (one image that is available for multiple architectures)
- Forked PR pipeline works the same (except with the hold step in between)
- Fixed the tag pipeline and build a squash image for a given tag
- Updated the Makefile to contain all the needed commands to run locally on M1 macs as well.
- Enhanced documentation about how to run things

---

image


lukasholzer:
Currently blocked by https://github.com/netlify/binrc/issues/36

.editorconfig
```diff
@@ -11,3 +11,6 @@ trim_trailing_whitespace = true
 insert_final_newline = true
 indent_style = space
 indent_size = 2
+
+[Makefile]
```
lukasholzer, line 15: Makefiles can only have tabs as indents

.github/workflows/whalescale.yml
```diff
@@ -1,25 +0,0 @@
-name: "Check docker image size"
```
lukasholzer, line 1: We don't want to build it twice on github (cost side and time side) so the benefit we get out of it is that minimal that we should delete it and once we have time we should build this for circleci where we already have built the images

Dockerfile
```diff
@@ -328,7 +330,7 @@ ENV BINRC_VERSION 0.2.9
 
 RUN mkdir /opt/binrc && cd /opt/binrc && \
     curl -sL https://github.com/netlify/binrc/releases/download/v${BINRC_VERSION}/binrc_${BINRC_VERSION}_Linux-64bit.tar.gz | tar zxvf - && \
-    ln -s /opt/binrc/binrc_${BINRC_VERSION}_linux_${TARGETARCH}/binrc_${BINRC_VERSION}_linux_${TARGETARCH} /usr/local/bin/binrc
+    ln -s /opt/binrc/binrc_${BINRC_VERSION}_linux_amd64/binrc_${BINRC_VERSION}_linux_amd64 /usr/local/bin/binrc
```
lukasholzer, line 333: The amd version works fine on arm as well 🤷

lukasholzer:
### Forked PR's

the onhold works fine

CleanShot 2022-08-19 at 11 40 15@2x

CleanShot 2022-08-19 at 13 45 42@2x

The image on hub.docker.com
CleanShot 2022-08-19 at 13 49 23@2x

lukasholzer:
The overall pipeline time went down from 20min to 14min by removing the need of saving and attaching the images and doing the testing and building in one stage

CleanShot 2022-08-19 at 14 57 39

lukasholzer:
working for tags as well: 

CleanShot 2022-08-26 at 11 53 54

https://hub.docker.com/layers/build/netlify/build/v4.10.1-test.7-squash/images/sha256-abfb143ef652cd3bfd6aef99dca7607bd83bab02352e83268d0e0a8ab52ca2d1?context=explore

CONTRIBUTING.md
```diff
@@ -5,6 +5,41 @@ please read the [code of conduct](CODE_OF_CONDUCT.md).
...
+
+#### Running the build
+
+If you are inside the image you need to run the build script:
+
+```bash
+/opt/build-bin/build
+```
+
+#### Running functions individually
```
aitkenster, line 24: ❤️ thanks for adding more detailed instructions here

Dockerfile
```diff
@@ -328,7 +333,7 @@ ENV BINRC_VERSION 0.2.9
...
```
aitkenster, line 337: Why doesn't `$TARGETARCH` need to be used here in the end?

lukasholzer: The amd64 binary is working on arm as well :D there is no arm64 binary – seems like a general linux version

see: https://github.com/netlify/binrc/releases/tag/v0.2.9

Dockerfile
```diff
@@ -343,10 +348,14 @@ USER root
 #
 ################################################################################
 
-ENV HUGO_VERSION 0.85.0
```
JGAntunes, line 346: I was trying to see if they follow semver and feels like they don't(?)
- https://github.com/gohugoio/hugo/releases

I guess there's no easy way of changing this default without uninteded impact but wondering if we could lower this by sharing some update (forum post?) on the matter?

JGAntunes (commented):
🚀 looking really good!

LGTM but left a single comment on the `HUGO_VERSION` change. I'm concerned this might be impactful for folks relying on the default version. Maybe there's some way we can easen this through comms? 🤔

Dockerfile
lukasholzer: Good point will reach out to support

kitop: Fun fact: The documented version is off with the actual version:
https://github.com/netlify/build-image/blob/focal/included_software.md
https://github.com/netlify/build-image/blob/focal/Dockerfile#L346

Besides that, I'm wondering if we have any data on hugo versions being used?
A simple forum update might work for support to point folks to if any issues.

Dockerfile
```diff
@@ -343,10 +348,14 @@ USER root
...
+ENV HUGO_VERSION 0.101.0
 
-RUN binrc install gohugoio/hugo ${HUGO_VERSION} -c /opt/buildhome/.binrc | xargs -n 1 -I{} ln -s {} /usr/local/bin/hugo_${HUGO_VERSION} && \
-    ln -s /usr/local/bin/hugo_${HUGO_VERSION} /usr/local/bin/hugo
+RUN case "$TARGETARCH" in \
+      "arm64") HUGO_FILE="hugo_${HUGO_VERSION}_Linux-ARM64.deb" ;; \
+      "amd64") HUGO_FILE="hugo_${HUGO_VERSION}_Linux-64bit.deb" ;; \
+    esac && \
+    wget -nv --quiet https://github.com/gohugoio/hugo/releases/download/v0.101.0/${HUGO_FILE} && \
```
kitop: Shouldn't the `HUGO_VERSION` be used instead of hardcoded `v.0.101.0` in the path?

```suggestion
    wget -nv --quiet https://github.com/gohugoio/hugo/releases/download/${HUGO_VERSION}/${HUGO_FILE} && \
```

Dockerfile
```diff
@@ -34,12 +33,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     echo 'LANGUAGE="en_US:en"' >> /etc/default/locale && \
     locale-gen en_US.UTF-8 && \
     update-locale en_US.UTF-8 && \
-    apt-key adv --fetch-keys https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc && \
     add-apt-repository -y ppa:ondrej/php && \
-    add-apt-repository -y ppa:openjdk-r/ppa && \
-    add-apt-repository -y ppa:git-core/ppa && \
-    add-apt-repository -y ppa:deadsnakes/ppa && \
```
kitop: Curious about this 2 repositories being removed. I see Erlang was moved, but curious about why removing git-code and deadsnakes (what even is that?)

lukasholzer: it seemed to make no difference so I removed them. The repositories are only used for apt-get installs and I think they where added for erlang.

As they are not used anymore (and the dockerfile is the only place for apt-get install) we should be safe to remove them.

Dockerfile
lukasholzer: I reverted to the original version to get this PR over the finishing line

lukasholzer:
@kitop can you try running `make build-base` on your local mac? I guess you have a intel one or?

Dockerfile
JGAntunes: deadsnakes packages python versions for ubuntu which historically lagged behind in the default repos of certain releases I believe - https://launchpad.net/%7Edeadsnakes/+archive/ubuntu/ppa. I don't think it's needed for focal any longer given the versions we're installing:
- https://packages.ubuntu.com/focal/python/

As for `git` though - https://launchpad.net/~git-core/+archive/ubuntu/ppa - we should double check the version we're installing by dropping this ppa (and compare to what we have right now). In the long run I also think we should probably pin this since it's a critical dep for our whole build pipeline.

kitop:
> @kitop can you try running `make build-base` on your local mac? I guess you have a intel one or?

@lukasholzer it seemed to work, took 815.1s. Do you want me to run anything else in particular? tests? a script?

lukasholzer:
> > @kitop can you try running `make build-base` on your local mac? I guess you have a intel one or?
> 
> @lukasholzer it seemed to work, took 815.1s. Do you want me to run anything else in particular? tests? a script?

nope just wanted to check if the local commands are working with make (gathering the correct archname on amd machines) Tests and everything are working on the CI so should be more than good to go

Dockerfile
lukasholzer: added the git core back for now but still don't think we need it as it is just the repository (and we are not doing anything with it.

JGAntunes (approved):
🚀

PR #829, opened 2022-08-16
feat: update nvm and yarn to latest version

danez:
#### Summary

yarn 1 isn't developed anymore, so it might make sense to use the latest version, because it has the best compatibility with newer versions of yarn.

nvm changelog doesn't seem too interesting, one thing that might be of interest is this line
>nvm_get_download_slug: better architecture selection for M1 Macs

which might help with amd64 build of the image.


JGAntunes:
Thanks for your contribution @danez 🙏

For yarn specifically, we pin that version at the API level for newly created sites. So we try to have these two version in synch. See here:
- https://github.com/netlify/bitballoon/blob/388344a48b002fd59ab3ba17d8bf4ee274d4c3fc/app/models/repo.rb#L23

danez:
Exactly, wanted to update bitballoon once this is done. But I can prepare a PR

danez:
Here is the bitballoon PR: https://github.com/netlify/bitballoon/pull/13187

JGAntunes (changes requested):
Looks like this test needs updating too (this probably became redundant with the container structure tests so we might consider drop it too I guess 🤔):
- https://github.com/netlify/build-image/blob/146f9d055346de28b685bd81ddf96fd599ee8360/tests/node/yarn.bats#L30-L33

JGAntunes:
> Here is the bitballoon PR: https://github.com/netlify/bitballoon/pull/13187

Thanks! I think once we go through merging and releasing this we can go through releasing the Bitballoon one.

danez:
I fixed it all hopefully.

JGAntunes (dismissed):
🚀

danez:
Only focal is good enough?

JGAntunes:
Yep, that should be enough @danez 👍

lukasholzer (approved):
LGTM!

PR #836, opened 2022-09-26
feat: add Deno to focal

Skn0tt:
This PR installs Deno on the Focal image.

related to https://github.com/netlify/pillar-workflow/issues/504-


Dockerfile
```diff
@@ -300,6 +300,22 @@ ENV PATH "/opt/buildhome/.yarn/bin:$PATH"
...
 USER root
 
+################################################################################
+#
+# Deno
+#
+################################################################################
+
+RUN if [ "$TARGETARCH" = "amd64" ] ; then curl -o- -L https://deno.land/x/install/install.sh > /usr/local/bin/deno-installer.sh; fi
+ENV DENO_VERSION=v1.25.4
```
eduardoboucas, line 310: Might be good to add a comment explaining that this env variable will be used by the `install.sh` script to determine which version to download?

Skn0tt: the line where it's used is two lines down, not sure if a comment is necessary. we could make this an `ARG` instead of `ENV` to make this more obvious

Skn0tt: `install.sh` itself contains no reference to `DENO_VERSION`, only to `$1`

eduardoboucas: Oh, sorry I totally misread this. Ignore me!

Skn0tt:
The failing test doesnt look related to this PR.

kitop:
PR fixing failing test: https://github.com/netlify/build-image/pull/837

You can rebase once that's shipped.

kitop (approved):
🦕

PR #837, opened 2022-09-27
fix: update test to reflect new homebrew version

kitop:
#### Summary

Fixes failing test in amd64.

Homebrew has released a new minor version, and as we're installing from HEAD we should reflect that in the tests.


lukasholzer (approved):
Thx so much for taking care of this 🐳

PR #843, opened 2022-10-07
feat: Install PHP 8.1

bakerkretzmar:
Adds PHP 8.1 to the build image. Just like 7.4, it can be selected and used in builds by setting the `PHP_VERSION` environment variable.

PHP 8.1 was released almost a year ago, and PHP 8.0 (the highest version currently included in this build image) will stop receiving active support next month when 8.2 is released.

See https://www.php.net/supported-versions.php

To test this I built and ran the image locally and verified that PHP 8.1 is installed and runs.

---

image


jobala (approved):

intrepidws:
As PHP 7.4 has been EOL'd and PHP 8.2 is now out, I wanted to check in on the status of a build with PHP 8.1. Are there any plans to release this?

kitop:
We're looking into this and aiming to release it soon!

PR #845, opened 2022-10-11
feat: add pnpm to build image

### LINKED ISSUE #789: Latest NPM versions throw an error when workspaces are configured
**Describe the bug**

Versions of NPM greater than 8.4 throw an error if the build is using NPM workspaces. It seems like the reason is workspaces aren't compatible with  [npm set](https://github.com/netlify/build-image/blob/a49edbc55fc722f80f6b85963a5060f8d043a844/run-build-functions.sh#L183) so we might need to rethink this 
command

**To Reproduce**

1. Create a monorepo using npm workspaces. 
2. Deploy to Netlify.
3. See the errors being logged.

**Configuration**

NPM_VERSION isn't set

**Expected behavior**

No error logs

**Build Output**

```
6:28:58 PM: npm workspaces detected
6:28:58 PM: Installing NPM modules using NPM version 8.11.0
6:28:59 PM: npm ERR! code ENOWORKSPACES
6:28:59 PM: npm ERR! This command does not support workspaces.
6:28:59 PM: npm ERR! A complete log of this run can be found in:
6:28:59 PM: npm ERR!     /opt/buildhome/.npm/_logs/2022-06-15T16_28_58_982Z-debug-0.log
```

**Additional context**

lukasholzer:

#### Summary

Fixes https://github.com/netlify/build/issues/1633
Fixes #789

Add pnpm support to the build image via the new corepack feature of node
as @ascorbic requested to have corepack enabled

Latest buildbot image with it: `e3838a70621b55be5e32e035882193d2a1a220a4-focal`

The matching Xenial PR that implements the breaking change in the `install_dependencies` function: https://github.com/netlify/build-image/pull/846

https://app.netlify.com/teams/lukas-holzer/builds/634931c9058c2417c7f14c5a#L34-L52
CleanShot 2022-10-14 at 11 56 51

if run with an too old node version we error out:
CleanShot 2022-10-14 at 12 00 16

---


Dockerfile
```diff
@@ -279,23 +279,26 @@ RUN curl -o- -L https://yarnpkg.com/install.sh > /usr/local/bin/yarn-installer.s
...
-ENV ELM_VERSION=0.19.1-5
 ENV YARN_VERSION=1.22.19
+ENV PNPM_VERSION=7.13.4
 
 ENV NETLIFY_NODE_VERSION="16"
 
 RUN /bin/bash -c ". ~/.nvm/nvm.sh && \
          nvm install --no-progress $NETLIFY_NODE_VERSION && \
          npm install -g grunt-cli bower && \
-             bash /usr/local/bin/yarn-installer.sh --version $YARN_VERSION && \
```
danez, line 297: I guess we could remove the step that downloads this script (line 278), because I guess it would conflict with corepack anyway? OR if we want to keep it, make our own version that uses corepack?

lukasholzer: Good point thx forgot that

run-build-functions.sh
```diff
@@ -177,10 +162,29 @@ run_yarn() {
   export PATH=$(yarn bin):$PATH
 }
 
-run_npm_set_temp() {
```
lukasholzer, line 180: This is not used anymore: 

CleanShot 2022-10-12 at 12 48 09

lukasholzer:
Tested it on a sample build: 
https://app.netlify.com/sites/build-reproductions/deploys/6346aff0f6f89370fa12710e#L30-L51

lukasholzer:
CleanShot 2022-10-12 at 14 20 39@2x

run-build-functions.sh
lukasholzer: This fixes: https://github.com/netlify/build-image/issues/789

kitop:
I tried this with yarn and an older node version (14.18.3) and it fails due to the lack of corepack: https://app.netlify.com/sites/eloquent-kringle-e5fd85/deploys/6346c5554bb47610f3b2d67f

CleanShot 2022-10-12 at 15 37 39@2x


This could break a lot of builds. How should we deal with this case?

run-build-functions.sh
```diff
@@ -132,32 +132,17 @@ restore_node_modules() {
...
 run_yarn() {
   yarn_version=$1
-  if [ -d $NETLIFY_CACHE_DIR/yarn ]
-  then
-    export PATH=$NETLIFY_CACHE_DIR/yarn/bin:$PATH
-  fi
   restore_home_cache ".yarn_cache" "yarn cache"
 
-  if [ $(which yarn) ] && [ "$(yarn --version)" != "$yarn_version" ]
+  if [ "$(yarn --version)" != "$yarn_version" ]
```
jobala: What happens if yarn doesn't exist?

lukasholzer: yarn is always installed on the system

ascorbic:
Yeah, I think this needs to ensure it's using a compatible version of node before trying to use corepack. It was added in v16.9.0 and v14.19.0

lukasholzer:
@kitop I'm currently questioning why it should be possible to run with a lower node minor version 🤔 If you specify node 14 compatible you should run always on the latest minor we support and not on an old version.

But other than that I would suggest reverting the install of yarn back to not use corepack and just install corepack and pnpm if your node version is greater equal than 14.19.0

run-build-functions.sh
JGAntunes: Albeit we install `yarn` by default, we do it for the given default node version within build image:
- https://github.com/netlify/build-image/blob/650ad5b24bef4bd1740bac32c70bb791722993bd/Dockerfile#L291-L298

I believe that by installing or sourcing a different node version configured by the customer, yarn won't be present?

JGAntunes:
>  I would suggest reverting the install of yarn back to not use corepack

Yeah I don't think we can tie yarn's usage with corepack without introducing a breaking change or going through extra trouble setting it up. We should leave outside of the scope of this change.

kitop:
Are you saying to only use corepack for pnpm and keep the old method for yarn? That'd be a good way to unblock, if there are no other tradeoffs.
We'll still need to support yarn for Node 12 and lower, anyway so this might simplify at the moment?

I've had a chat with Lukas about the node version and using latest minor, and that could be it's own thing.

lukasholzer:
@kitop and @JGAntunes changed the installing

lukasholzer:
Now we are using by default corepack and have yarn + pnpm installed over corepack.

When a user decides to use a old node version (where corepack is not supported) we install yarn then with the old method. 

If the user uses a newer node version we can switch the version with corepack

This was required as having corepack installed + trying always to use the old way to install yarn does not work (you cannot uninstall yarn then if corepack is enabled and yarn got installed even over the old method)

Dockerfile
```diff
@@ -274,29 +274,31 @@ USER root
...
-         nvm alias default node && nvm cache clear"
-ENV PATH "/opt/buildhome/.yarn/bin:$PATH"
+         nvm alias default node && \
+         nvm cache clear && \
+         corepack enable && \
+         corepack prepare yarn@$YARN_VERSION --activate && \
```
lukasholzer, line 300: per default we install already yarn + pnpm with corepack

run-build-functions.sh
```diff
@@ -132,32 +139,38 @@ restore_node_modules() {
...
-  fi
   restore_home_cache ".yarn_cache" "yarn cache"
 
-  if [ $(which yarn) ] && [ "$(yarn --version)" != "$yarn_version" ]
-  then
-    echo "Found yarn version ($(yarn --version)) that doesn't match expected ($yarn_version)"
-    rm -rf $NETLIFY_CACHE_DIR/yarn $HOME/.yarn
-    npm uninstall yarn -g
-  fi
+  if ! [ $(which corepack) ]; then
```
lukasholzer, line 137: if corepack is not available (then the user has switched to a different - older - node version and we have to install yarn with the old way

run-build-functions.sh
```diff
@@ -219,46 +260,9 @@ run_npm() {
...
-    check_python_version "runtime.txt"
-  elif [ -f Pipfile ]
-  then
-    echo "Found Pipfile restoring Pipenv virtualenv"
-    restore_cwd_cache ".venv" "python virtualenv"
-  else
-    PYTHON_VERSION=$defaultPythonVersion
-    check_python_version "the PYTHON_VERSION environment variable"
-  fi
+install_node() {
```
lukasholzer, line 256: moved that to it's own function to have it better testable

lukasholzer:
@kitop 

on old node version it works with yarn fine as well:
CleanShot 2022-10-14 at 12 03 47

kitop:
Tried a few cases and all working for me! Would love some extra 👀 on the code tho

run-build-functions.sh
```diff
@@ -72,6 +72,13 @@ mkdir -p $NETLIFY_CACHE_DIR/.cargo
 : ${NPM_FLAGS=""}
 : ${BUNDLER_FLAGS=""}
 
+# get's the major version out of a string
+get_major_version() {
+  local version=$1
+  # The sed replaces all non alphanumeric values if a version starts with `v1.3.0` it should provide `1.4.0`
```
aitkenster:
```suggestion
  # The sed replaces all non alphanumeric values if a version starts with `v1.4.0` it should provide `1.4.0`
```

run-build-functions.sh
```diff
@@ -177,10 +190,38 @@ run_yarn() {
...
-  npm set tmp $HOME/tmp
+
+run_pnpm() {
+  pnpm_version=$1
+  restore_home_cache ".pnpm-store" "pnpm cache"
+
+  if ! [ $(which corepack) ]; then
+    echo "Error while installing PNPM $pnpm_version"
+    echo "We cannot install the expected version of PNPM ($pnpm_version) as your required Node.js version $NODE_VERSION does not allow that"
+    echo "Please assure that you use at least Node Version 14.19.0 or greater than 16.9.0"
```
aitkenster:
```suggestion
    echo "Please ensure that you use at least Node Version 14.19.0 or greater than 16.9.0"
```

run-build-functions.sh
```diff
@@ -296,13 +300,73 @@ install_dependencies() {
...
   fi
 
+  # remove the v in front of the node version
+  local cleanNodeVersion=$(echo "$NODE_VERSION"| sed "s/[^[:digit:].-]//g")
+  # Node.js Corepack
+  # corepack was packported to 14.19.0
+  dpkg --compare-versions "$cleanNodeVersion" "ge" "14.19.0"
+
+  if [ $? -eq 0 ]; then
+    # corepack was added in 16.9.0 and packported to 14.19.0
```
aitkenster:
```suggestion
    # corepack was added in 16.9.0 and backported to 14.19.0
```

run-build-functions.sh
JGAntunes:
```suggestion
  # The sed replaces all non alphanumeric values if a version starts with `v1.4.0` it should provide `1.4.0`
```

run-build-functions.sh
```diff
@@ -296,13 +300,73 @@ install_dependencies() {
...
+    dpkg --compare-versions "$cleanNodeVersion" "ge" "16.9.0"
+    if [ $? -eq 0 ] || [ "$(get_major_version $cleanNodeVersion)" == "14" ]; then
+      echo "Enabling node corepack"
+      corepack enable
+    fi
+  fi
```
JGAntunes: Would it be possible to just check via `which corepack` or run a `corepack` command and via its error code detect if corepack is present or not and log it accordingly? Thinking it would be easier and would avoid the version comparison and detection/

aitkenster (dismissed):
LGTM! Just spotted a few typos

run-build-functions.sh
```diff
@@ -132,32 +139,38 @@ restore_node_modules() {
...
+      bash /usr/local/bin/yarn-installer.sh --version $yarn_version
+      mv $HOME/.yarn $NETLIFY_CACHE_DIR/yarn
+      export PATH=$NETLIFY_CACHE_DIR/yarn/bin:$PATH
+    fi
+  else
+    # if corepack is installed use it for changing the yarn version
+    if [ "$(yarn --version)" != "$yarn_version" ]; then
+      echo "Installing yarn at version $yarn_version"
+      corepack prepare yarn@$yarn_version --activate
+    fi
```
JGAntunes, line 161: Could be farteched but I'm thinking on the following scenario:
- Someone running node configured via `NODE_VERSION`, `v18.x` for example (they have corepack).
- This change lands in prod
- `which corepack` will return a non-error exit code
- `yarn` binary won't exist so `yarn --version` will blow up? 🤔
Would this be possible? Should we check for the presence of the yarn executable before too? If so maybe we could move the `which yarn` assertion to the top and use that across the function via a `yarn_exists` flag maybe?

JGAntunes (commented):
Left a couple of suggestions and an edge case which I'm not entirely sure is possible or not but thought would be good to raise 😅

run-build-functions.sh
lukasholzer: That's not the case corepack installs yarn for each node version which is super awesome:

you can test it by: 

1. current node version - uninstall your current yarn binary
2. run `which yarn` (should fail)
3. run `corepack enable`
4. run `corepack prepare yarn@1.22.19 --activate`
5. run `which yarn` should work
6. run `nvm install 18` (has corepack but no yarn installed)
7. run `which yarn` (should fail)
8. run `corepack enable` 
9. run `which yarn` should work 🥳 

I have added an extra test case testing this scenario :)

run-build-functions.sh
lukasholzer: sure! Good point (was kind of a relict from an earlier version of the PR)

run-build-functions.sh
JGAntunes: Ah nvm, I was thinking about the scenario where your node installation was previously cached, however we always go through the `install_node` function which does the `corepack enable` step 👍

JGAntunes (commented):
Just an unused function leftover I think. Other than that LGTM 👍

lukasholzer:
> Just an unused function leftover I think. Other than that LGTM 👍

oh yea missed that already deleted

JGAntunes (approved):

PR #846, opened 2022-10-12
fix: incorporate breaking change in install_dependencies order

lukasholzer:

#### Summary

This is the matching Xenial PR that just incorporates the breaking change in the order of the install_dependency function.

We won't add pnpm support to xenial anymore as this is already in the deprecation process
The focal PR: https://github.com/netlify/build-image/pull/845

---


run-build.sh
```diff
@@ -21,11 +21,12 @@ cd $NETLIFY_REPO_DIR
 : ${NODE_VERSION="12.18.0"}
 : ${RUBY_VERSION="2.7.1"}
 : ${YARN_VERSION="1.22.4"}
+: ${PNPM_VERSION="1.22.4"}
```
danez: Should that be 7.13.4? Not that it matters, but...

lukasholzer: you are right – stupid typo but does not matter (for sake of correctness I will update it)

jobala (approved):

PR #854, opened 2022-10-19
chore: run pre & post install scripts

### LINKED ISSUE #523: npm preinstall and postinstall scripts are not run
npm preinstall and postinstall scripts are not run, resulting in breaking builds that are valid outside of Netlify.

This is caused by our caching mechanism only restoring the `node_modules/` directory:

https://github.com/netlify/build-image/blob/d84c79427e8f83c1ba17bcdd7b3fe38059376b68/run-build-functions.sh#L478

And then running npm install conditionally:

https://github.com/netlify/build-image/blob/d84c79427e8f83c1ba17bcdd7b3fe38059376b68/run-build-functions.sh#L149-L153

An existing PR fixes this: #510.

jobala:
#### Summary

Fixes #523 

Screenshot 2022-10-24 at 12 50 01

Removes the cache check and always runs `npm install`
---


lukasholzer (commented):
Looks good from looking at it. Can you maybe write a bats test for it? especially the thing with the feature flag?

run-build-functions.sh
```diff
@@ -237,20 +238,34 @@ run_npm() {
     fi
   fi
 
-  if install_deps package.json $NODE_VERSION $NETLIFY_CACHE_DIR/package-sha
+  if has_feature_flag "$featureFlags" "buildbot_bypass_module_cache"
   then
     echo "Installing NPM modules using NPM version $(npm --version)"
-
-    if npm install ${NPM_FLAGS:+$NPM_FLAGS}
+    if npm install ${NPM_FLAGS:+"$NPM_FLAGS"}
```
JGAntunes, line 245: I'm wondering if it would make sense to add something in the logs for this scenario, so that is more easily identifiable + make it easier to aggregate and compare on possible impacts of this fix as we roll it out?

tests/node/bypass_cache.bats
```diff
@@ -0,0 +1,32 @@
...
+  load '../../run-build-functions.sh'
+}
+
+teardown() {
+  rm -rf $NETLIFY_CACHE_DIR/package-sha
+}
+
+@test 'bypass_cache run pre and post install scripts' {
+  run run_npm "buildbot_bypass_module_cache"
+  assert_output --partial "NPM modules installed"
```
JGAntunes, line 26: Should we also:
```
refute_output --partial Creating package sha"
```

JGAntunes (commented):
Looking good 👍 left a small set of comments

JGAntunes (approved):
🎉

PR #870, opened 2022-11-15
fix: Bump Go to 1.19

danez:
#### Summary

What I noticed is that we install the latest patch version with `1.19.x` but then always use and also freeze (in bitballoon) 1.19 version. So we never actually use the latest patch version. Should we also freeze and use `1.19.x` by default? That though would mean that new sites that have `1.19.x` frozen will automatically get patch updates installed.

If we freeze 1.19 we should also install 1.19 and not the latest patch version of 1.19.x

Ref: netlify/bitballoon#13837

Fixes: https://github.com/netlify/pod-compute/issues/188

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -458,7 +458,7 @@ ENV PATH "$PATH:/opt/buildhome/.gimme/bin"
 ENV GOPATH "/opt/buildhome/.gimme_cache/gopath"
 ENV GOCACHE "/opt/buildhome/.gimme_cache/gocache"
 # Install the default version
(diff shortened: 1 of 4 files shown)


lukasholzer (dismissed):

jobala (dismissed):

JGAntunes (dismissed):
Agree we should change the `freezed` version to `1.19.x`, I believe that has likely been a leftover and folks probably assumed `1.19` would result in `1.19.x`.

danez:
Updated to use `1.19.x` as default version

jobala (approved):

PR #887, opened 2022-12-17
fix: ensure that go is always correctly installed

danez:
#### Summary

Fixes https://github.com/netlify/pillar-workflow/issues/1025

When installing go we always checked if `GIMME_GO_VERSION` (default installed version in docker `1.19.x`) does not equal `installGoVersion` (users version from `GO_VERSION` or default 1.19.x).

If they do not equal we would resolve `installGoVersion` and install it by calling gimme.

Otherwise we would install the version from `GIMME_GO_VERSION`.

Checking the history `GIMME_GO_VERSION` was always identical to the default `installGoVersion` unless the user changed it, but in #870 both default value were changed to be `1.19.x`. Two problems with that:

- We tried to source the env file with `GIMME_GO_VERSION` (the unresolved version), but gimme always resolves before installing, so there is no env file for 1.19.x (this seems to have been broken since #870)
- Now that 1.19.4 was released, 1.19.x resolves to it, but in the build-image we have 1.19.3 installed.

The change simplifies the code and always checks if the actual resolved version from `installGoVersion` is already installed or not. GIMME_GO_VERSION is not used anymore as it really does not matter, either the requested go version is installed or not.

This has one downside though: we always call `gimme --resolve` which goes to network for `.x` versions, but caches known versions for 3 hours by default, I changed that to 7 days, but that is debatable. That means that when `1.19.5` is released it might take up to a week before users with a go version of `1.19.x` will get that new version. ([The actual check in gimme](https://github.com/travis-ci/gimme/blob/master/gimme#L689) uses the file creation time, so I do hope our caching preserves file creation times)

~~What I did not understand what `gimme | bash` was supposed to do? It installs GIMME_GO_VERSION by default, but why the pipe to bash?~~ gimme returns a bash script, so the `source` call we had in the script did source the script a second time.

The alternative is to simply revert all the default versions from 1.19.x to 1.19.4, BUT bitballon currently freezes 1.19.x to new user accounts, so in any way the builds would get a `installGoVersion` of 1.19.x

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -468,6 +468,8 @@ ENV GOCACHE "/opt/buildhome/.gimme_cache/gocache"
 # Install the default version
 ENV GIMME_GO_VERSION "1.19.x"
 ENV GIMME_ENV_PREFIX "/opt/buildhome/.gimme/env"
+ENV GIMME_VERSION_PREFIX "/opt/buildhome/.gimme/versions"
+ENV GIMME_TYPE "binary"
 RUN gimme | bash
 
 ################################################################################
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -1014,8 +1014,15 @@ install_go() {
       installGoVersion="$goVersion"
     fi
   fi
+
   # Cache known version for 7days (604800 seconds)
-  resolvedGoVersion=$(GIMME_KNOWN_CACHE_MAX=604800 gimme --resolve $installGoVersion)
+  resolvedGoVersion="$(GIMME_KNOWN_CACHE_MAX=604800 gimme --resolve $installGoVersion)"
+  if [ $? -ne 0 ]
+  then
+    echo "Failed to resolve Go version '$installGoVersion'"
+    exit 1
+  fi
+
   gimmeEnvFile=$HOME/.gimme/env/go$resolvedGoVersion.linux.$(dpkg --print-architecture).env
 
   # Check if the version is already installed by gimme
--- a/run-build-functions.sh
+++ b/run-build-functions.sh
@@ -1014,29 +1014,21 @@ install_go() {
       installGoVersion="$goVersion"
     fi
   fi
+  # Cache known version for 7days (604800 seconds)
+  resolvedGoVersion=$(GIMME_KNOWN_CACHE_MAX=604800 gimme --resolve $installGoVersion)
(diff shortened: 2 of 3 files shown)


jobala (dismissed):

Dockerfile
```diff
@@ -468,6 +468,11 @@ ENV GOCACHE "/opt/buildhome/.gimme_cache/gocache"
 # Install the default version
 ENV GIMME_GO_VERSION "1.19.x"
 ENV GIMME_ENV_PREFIX "/opt/buildhome/.gimme/env"
+ENV GIMME_VERSION_PREFIX "/opt/buildhome/.gimme/versions"
+# Default is auto which on invalid go versions starts cloning git repos and searching through git and tries to build from git
+ENV GIMME_TYPE "binary"
```
danez, line 473: This will not allow installs from git anymore. Previously one could specify a git commit and gimme would clone the repo and start building.
When we do `gimme --resolve something` with an invalid version it will also clone the go repo and start searching if there is a commit that matches.

Dockerfile
```diff
@@ -468,6 +468,8 @@ ENV GOCACHE "/opt/buildhome/.gimme_cache/gocache"
 # Install the default version
 ENV GIMME_GO_VERSION "1.19.x"
 ENV GIMME_ENV_PREFIX "/opt/buildhome/.gimme/env"
+ENV GIMME_VERSION_PREFIX "/opt/buildhome/.gimme/versions"
+ENV GIMME_TYPE "binary"
```
danez, line 472: ensure we always install a binary version when building the docker image
and set the same paths as the install script.

jobala (approved):

run-build-functions.sh
```diff
@@ -1015,28 +1015,27 @@ install_go() {
     fi
   fi
 
-  if [ "$GIMME_GO_VERSION" != "$installGoVersion" ]
+  # Cache known version for 7days (604800 seconds)
+  resolvedGoVersion="$(GIMME_KNOWN_CACHE_MAX=604800 gimme --resolve $installGoVersion)"
```
danez, line 1019: I wonder if doing this now always can impact performance or even error rate. This needs network and then caches the results for 7 days.

If a real version is supplied like 1.19 or 1.19.3 this will not access network. Only if the `installGoVersion ` is `stable` or `1.19.x` which is the default.

